PR #5, opened 2019-05-06
CIF-827 - Fix NPE in product model for getDescription

herzog31:
## Description
Fixed NullPointerException that might occur when calling `getDescription` for a product or product variant.

```
Caused by: java.lang.NullPointerException: null
	at com.adobe.cq.commerce.core.components.internal.models.v1.product.ProductImpl.getDescription(ProductImpl.java:122) [com.adobe.cq.commerce.components.core:0.0.1.SNAPSHOT]
```

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/ProductImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/ProductImpl.java
@@ -27,6 +27,7 @@ import javax.inject.Inject;
 import com.adobe.cq.commerce.graphql.client.GraphqlClient;
 import com.adobe.cq.commerce.graphql.client.GraphqlRequest;
 import com.adobe.cq.commerce.graphql.client.GraphqlResponse;
+import com.adobe.cq.commerce.magento.graphql.ComplexTextValue;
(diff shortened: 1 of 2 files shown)


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/ProductImpl.java
```diff
@@ -395,4 +396,18 @@ private ProductInterface getProduct(GraphqlClient client, String slug) {
...
+
+    private String safeDescription(ProductInterface product) {
+        ComplexTextValue description = product.getDescription();
+        if (description == null) {
+            return "";
+        }
+
+        String descriptionHtml = description.getHtml();
+        if (descriptionHtml == null) {
+            return "";
```
cjelger: Why return an empty string? The HTL should handle that correctly, simply testing for the property before displaying anything.

herzog31: That's probably the easier approach. So I let the function just return `null`.

herzog31: @cjelger done :)

PR #7, opened 2019-05-07
releng - Update jackson dependency to fix vulnerability

herzog31:


cjelger:
Did you check that these versions are ok for AEM 6.4.4? I'm actually wondering where we use these `jackson` dependencies in our code ... do we really need that other than for testing?

herzog31:
@cjelger The product component uses it to convert data objects into JSON to provide them to the client-side code via `data` attributes in the HTML.
I will verify that everything still works with AEM 6.4.

herzog31:
@cjelger As discussed, I removed the jackson dependencies completely, since they are included in the 6.4.4 UberJar. I also tried the all package on 6.4.0 and 6.4.4. It will not work with 6.4.0, but works on 6.4.4. However, 6.4.4. comes with a core components version that's too old, so a newer one has to be manually installed. I think we should declare this dependency explicitly in our pom file.

PR #10, opened 2019-05-08
@releng - updated README

mhaack:
## Description

Add missing configuration instructions

## Types of changes

## Checklist:


herzog31:
This project doesn't have a dependency to the authoring tooling yet (only GraphQL client + Magento models), so the picker will not be available. Either we add the dependency and update the documentation with the setup, or not mention the picker for now. But I guess for components like the product teaser we anyway need the dependency.

mhaack:
@herzog31 good point I will mention the authoring toolkit. I think we have to because the components already using the `cifcategoryfield` & `cifproductfield` in the dialogs and therefor have indirect dependencies.

cjelger:
Well I wouldn't call it an "indirect" dependency, in this case, it's really required. Maybe, but I'm really not sure here, we could also embed these dependencies in the "all" package.

mhaack:
Just updated the PR and named it "optional" as the components itself will work, but will be limited. Maybe that is enough? We can refine if we have other components like teaser ready which really don't work without it.

mhaack:
@cjelger for the all package I would not embed it directly since it is a but we can add it as a package dependency.

README.md
```diff
@@ -25,14 +27,40 @@ For more details about the UberJar please head over to the
...
+        * Keep the `default` service identifier or set something custom. Make sure to use the same value in step 2) below.
+        * For _GraphQL Service URL_ enter the URL of your Magento GraphQL endpoint (usually `https://hostname/graphql`)
+        * If you use this project together with the [AEM Commerce connector for Magento](https://github.com/adobe/commerce-cif-connector) this has to be configured only once.
+
+2) Link your AEM site with the GraphQL client instance
+    * Go to http://localhost:4502/crx/de/index.jsp
+    * Navigate to the page root node e.g. /content/venia/jcr:content for the Venia sample page
+    * Add a property `cq:graphqlClient` with the value of the service identifier of step 1
+    
+3) Assign the Magento root category of to your website
```
cjelger: typeo: "of to"

README.md
```diff
@@ -25,14 +27,40 @@ For more details about the UberJar please head over to the
...
+    * Go to AEM Sites console http://localhost:4502/sites.html/
+    * Navigate to the catalog page with your site structure e.g. /content/venia/us/en/products for the Venia sample page
+    * Select the catalog page and open page properties
+    * Select the Commerce tab
+    * Enter the root category of your Magento catalog the site should be assigned or use the category picker
```
cjelger: Sorry one more comment: maybe you should write "root category id", and I guess here the picker will not work without the connector, right? Maybe add an extra comment in parentheses?

mhaack: Changed the wording. The picker can, at the moment, only be used to select any category except the root category.

PR #19, opened 2019-05-13
CIF-722 - Remove obsolete properties

herzog31:
## Description
Updated feedback from verification. 
* Remove obsolete `catalogRoot` property.
* Fix `jcr:language` language tag.

For reference: The refsquad is recommending to use the Page API to get the language (https://git.corp.adobe.com/CQ/wcm/blob/master/bundles/cq-wcm-api/src/main/java/com/day/cq/wcm/api/Page.java#L357).

## Types of changes

## Checklist:

### DIFF
--- a/ui.content/src/main/content/jcr_root/content/venia/language-masters/en/products/.content.xml
+++ b/ui.content/src/main/content/jcr_root/content/venia/language-masters/en/products/.content.xml
@@ -6,7 +6,6 @@
         jcr:primaryType="cq:PageContent"
         jcr:title="Catalog Page"
         sling:resourceType="venia/components/structure/catalogpage/v1/catalogpage"
-        catalogRoot="{Boolean}true"
         showMainCategories="{Boolean}true"/>
     <category-page/>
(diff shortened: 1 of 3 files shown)


mhaack:
@herzog31 how to read your comment on the page API for language? What are you referring to, we already use the page API, see https://github.com/adobe/aem-core-cif-components/blob/453ac60c4f29b197cb89b21a6a69a1bdeccc2f87/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListItemImpl.java for example.

herzog31:
@mhaack Correct. Levente was suggesting to use `LanguageManager` in the JIRA issue.

LSantha:
LanguageManager is used in AEM in locations outside commerce. Ideally we would need to ensure that the available not deprecated methods used in AEM in our code or elsewhere to determine the language should yield consistent result. However at the current state of the affair that looks quite a challenging problem. We may need to revisit this issue in the future.

PR #20, opened 2019-05-14
CIF-632 - Support Magento multi store setups in GraphQL client

cjelger:
Adds Magento multi-store support

## Related Issue

CIF-632

## How Has This Been Tested?

Added unit test.

## Types of changes

## Checklist:


LSantha:
@cjelger , in an ideal world the models  would get a preconfigured GraphqlClient instance and the GraphqlClientWrapper would be an implementation detail behind the scenes. If we would have the adapter factory in this bundle that would be achievable or the cq:magentoStore could be handled in the graphql client bundle.

cjelger:
This doesn't really work because:
- In a multi-store setup, components could get a different `GraphqClient` instance depending on the `cq:graphqlClient` property which would be different for example for `/us/en` and `/ch/de`.
- We cannot handle `cq:magentoStore` in the graphql client bundle because this is not something generic, unless we add a property like `cq:graphqlClientHeaders` or something like that.

Note that we could add an adapter for `GraphqlClientWrapper`, but I don't think it would bring any extra benefit here. If you prefer that components use a `GraphqlClient` instead of a `GraphqlClientWrapper` (so we don't "expose" this class), we can easily change that.

LSantha:
@cjelger , I don't really see what is the impediment here. 
Now I tend to think that this Magento store view property could be handled very well in the Magento client. After all, there is always a store view in effect when we access Magento either the default or an other one. They could have added it to the endpoint URL. So that fact the it's in the header is not relevant. We want to specify the store view and not some arbitrary HTTP request headers.

cjelger:
But Levente ... we don't have a Magento client! The GraphQL client is generic, it doesn't have any Magento specific feature. That's exactly why I created that `GraphqlClientWrapper`, which after all, I'd could well rename to `MagentoGraphqlClient`.

LSantha:
@cjelger,
 
> The GraphQL client is generic, it doesn't have any Magento specific feature.

Indeed, thanks for the reminder. :-)

Nevertheless, there are a number of issues with this PR. Let me note that via in-line comments and I also have an idea for a solution.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImpl.java
```diff
@@ -179,7 +171,7 @@ private CategoryTreeQueryDefinition generateProductListQuery() {
      * @param categoryId the category id of category we request
      * @return {@link CategoryInterface}
      */
-    private CategoryInterface getCategory(GraphqlClient client, int categoryId) {
```
LSantha: Parameter 'client' is removed and JavaDoc is obsolete.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/MagentoGraphqlClient.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public static final String STORE_CODE_PROPERTY = "cq:magentoStore";
+
+    private GraphqlClient graphqlClient;
+    private RequestOptions requestOptions;
+
+    public MagentoGraphqlClient(Resource resource) {
+        graphqlClient = resource.adaptTo(GraphqlClient.class);
+        if (graphqlClient == null) {
+            LOGGER.warn("GraphQL client not available for resource " + resource.getPath());
```
LSantha: In this case this instance is invalid but the caller doesn't know about it. So they continue using it.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/MagentoGraphqlClient.java
```diff
@@ -0,0 +1,101 @@
...
+        String storeCode = properties.getInherited(STORE_CODE_PROPERTY, String.class);
+        if (storeCode != null) {
+            Header storeHeader = new BasicHeader("Store", storeCode);
+            requestOptions.withHeaders(Collections.singletonList(storeHeader));
+        }
+    }
+
+    @Override
+    public String getIdentifier() {
+        return graphqlClient.getIdentifier();
```
LSantha: NPE if graphqlClient is null.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/MagentoGraphqlClient.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * <b>This method uses the {@link RequestOptions} set by the MagentoGraphqlClient class.</b>
+     * <br/>
+     * <br/>
+     * {@inheritDoc}
+     */
+    @Override
+    public <T, U> GraphqlResponse<T, U> execute(GraphqlRequest request, Type typeOfT, Type typeOfU) {
+        return graphqlClient.execute(request, typeOfT, typeOfU, requestOptions);
```
LSantha: NPE if graphqlClient is null.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/MagentoGraphqlClient.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * <b>Only use this method if you do NOT want to use the {@link RequestOptions} set by the MagentoGraphqlClient class.</b>
+     * <br/>
+     * <br/>
+     * {@inheritDoc}
+     */
+    @Override
+    public <T, U> GraphqlResponse<T, U> execute(GraphqlRequest request, Type typeOfT, Type typeOfU, RequestOptions options) {
+        return graphqlClient.execute(request, typeOfT, typeOfU, options);
```
LSantha: NPE if graphqlClient is null.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/navigation/GraphQLCategoryProvider.java
```diff
@@ -18,54 +18,54 @@
...
-    private GraphqlClient graphqlClient;
+
+    private MagentoGraphqlClient client;
 
     GraphQLCategoryProvider(Page page) {
-        this.graphqlClient = page.getContentResource().adaptTo(GraphqlClient.class);
-        if (graphqlClient == null) {
-            LOGGER.warn("GraphQL client not available for page: " + page.getPath());
+        client = new MagentoGraphqlClient(page.getContentResource());
+        if (client == null) {
```
LSantha: This condition is always false. Even in the situation when your MagentoGraphqlClient instance is invalid (its internal GraphqlClient is null) and the old code would detect client == null and act accordingly. 
So, the error handling logic is invalid and you will get random NPE-s later.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/navigation/GraphQLCategoryProvider.java
```diff
@@ -18,54 +18,54 @@
...
+            LOGGER.warn("GraphQL client not available for resource {}", page.getContentResource().getPath());
         }
     }
 
     List<CategoryTree> getChildCategories(Integer categoryId) {
-        if (graphqlClient == null || categoryId == null) {
+        if (client == null || categoryId == null) {
```
LSantha: client is never null

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/navigation/GraphQLCategoryProvider.java
```diff
@@ -18,54 +18,54 @@
...
             return Collections.emptyList();
         }
 
         QueryQuery.CategoryArgumentsDefinition searchArgs = q -> q.id(categoryId);
         String queryString = Operations.query(query -> query.category(searchArgs, categoryQueryDefinition)).toString();
-        GraphqlResponse<Query, Error> response = graphqlClient.execute(new GraphqlRequest(queryString), Query.class, Error.class, QueryDeserializer.getGson());
+        GraphqlResponse<Query, Error> response = client.execute(new GraphqlRequest(queryString), Query.class, Error.class);
```
LSantha: Possible NPE thrown from execute() method.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/ProductImpl.java
```diff
@@ -86,86 +87,83 @@
...
 
-        // Get GraphqlClient from the page resource. This will look for a
-        // cq:catalogIdentifier property on the page or any ancestor and a
-        // corresponding OSGi configuration.
-        GraphqlClient client = resource.adaptTo(GraphqlClient.class);
+        // Get GraphqlClient from the resource.
+        client = new MagentoGraphqlClient(resource);
         if (client == null) {
-            LOG.error("Cannot get a GraphqlClient using the resource at {}", resource.getPath());
+            LOGGER.warn("GraphQL client not available for resource {}", resource.getPath());
```
LSantha: Unreachable code with all the consequences mentioned above.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImpl.java
```diff
@@ -97,19 +96,15 @@ private void initModel() {
 
         // get GraphQL client and query data
         if (categoryId != null) {
-            GraphqlClient client = resource.adaptTo(GraphqlClient.class);
-            this.category = getCategory(client, categoryId);
+            client = new MagentoGraphqlClient(resource);
```
LSantha: Older bug: missing error handling when client is null. 
Now client is never null but it can be invalid.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productteaser/ProductTeaserImpl.java
```diff
@@ -67,76 +78,77 @@ private void initModel() {
         }
         String productPath = properties.get(PRODUCT_PATH_PROP, String.class);
         if (productPath != null && !productPath.isEmpty()) {
-            String sku = getSkuFromUrl(productPath);
-            GraphqlClient client = resource.adaptTo(GraphqlClient.class);
+            String sku = getSkuFromPath(productPath);
+
+            // Get GraphqlClient from the resource.
+            client = new MagentoGraphqlClient(resource);
             if (client == null) {
```
LSantha: Client is never null + all the consequences.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/searchresults/SearchResultsImpl.java
```diff
@@ -74,20 +74,19 @@
...
+    private MagentoGraphqlClient client;
 
     Page productPage;
 
     @PostConstruct
     protected void initModel() {
         searchTerm = request.getParameter("search_query");
-        client = resource.adaptTo(GraphqlClient.class);
+        client = new MagentoGraphqlClient(resource);
         if (client == null) {
```
LSantha: Client is never null + all the consequences.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/MagentoGraphqlClient.java
cjelger: Fully agree, I left this as it was in all the existing sling models, but I agree it's much better to throw an exception. I initially changed it and then thought I leave it as it was. I will change that.

LSantha:
@cjelger, I propose to make the MagentoGraphqlClient an interface that we expose to our code and to customers code via adaptTo and keep its implementation private in the bundle. This is because adaptTo is a superior way of handling this dependency than a constructor. It fits better in the existing code and when adaptTo returns null you can more easily handle the null value than an exception. 
I propose to implement the adaption of Resource and Page to MagentoGraphqlClient.

I also propose to make MagentoGraphqlClient an independent interface from GraphqlClient with specialized methods that fits our use-cases. The existing calls to GraphqlClient.execute() all look the same: 

GraphqlClient.execute(new GraphqlRequest(queryString), Query.class, Error.class)

In our MagentoGraphqlClientImpl we can mostly likely handle the Error.class and probably the Query.class too or add special method for mutation when needed. This class then will handle the store view consistently and whatever else will come up. We can also think about new ways to make MagentoGraphqlClient more suitable for ourselves and our customers. GraphqlClient as is just requires too much boilerplate from the user. So MagentoGraphqlClient will be our recommendation for interacting with Magento in the server-side components and if that doesn't suite a customer they can still use GraphqlClient with all the ceremony.

cjelger:
@LSantha I included all your proposed changes except yet the interface and the adapter factory. I think we should first all agree that we want to export that, and then we can create a dedicated issue and another PR.

LSantha:
@cjelger 
> I included all your proposed changes except yet the interface and the adapter factory. I think we should first all agree that we want to export that, and then we can create a dedicated issue and another PR.

I think the answer is pretty obvious here:
* we have a specialized client now to interact with Magento and this is what we recommend to customers, so we need to export it
* we already agreed that the client taking its configuration from a resource adaptTo is a suitable approach
* should I also explain the interface?

Under these circumstances I wonder what are the pros of the static factory method?

cjelger:
I just think it's nice to agree with the team, so I propose that we create a separate issue, groom it, and then implement it. It will also require more unit tests for the adaptation, so it's quite a PR of its own.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/MagentoGraphqlClient.java
```diff
@@ -0,0 +1,101 @@
...
+     * This method returns <code>null</code> if the client cannot be instantiated.
+     * 
+     * @param resource The JCR resource to use to adapt to the lower-level {@link GraphqlClient}.
+     * @return A new MagentoGraphqlClient instance.
+     */
+    public static MagentoGraphqlClient create(Resource resource) {
+        try {
+            return new MagentoGraphqlClient(resource);
+        } catch (Exception e) {
+            LOGGER.error(e.getMessage());
```
LSantha: I don't think this should be an error. It may look like an error for the create method but not for an adaptTo. I prefer logging a warning on the call site if I don't get what I expect or I don't log anything if I decide so in any case I would live it up to the client to handle null as they want. 
By returning null I did my job, and that's also in the JavaDoc. There's no need to log anything here.

cjelger: ok I disagree, how do we resolve that? Did you decide to sabotage this PR?

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/MagentoGraphqlClient.java
```diff
@@ -0,0 +1,101 @@
...
+            return null;
+        }
+    }
+
+    private MagentoGraphqlClient(Resource resource) {
+        graphqlClient = resource.adaptTo(GraphqlClient.class);
```
LSantha: How about moving this logic outside of the constructor and passing the needed values as parameters? Then there is no need to throw an exception.

cjelger:
@LSantha I propose that you open a new PR to fix everything that you don't like.

PR #21, opened 2019-05-14
CIF-816 - Venia Logo in Header should link back to home page

LSantha:
## Description

 * navRoot property flags navigation root page
 * linked Venia Logo to navigation root page
 * unit tests

## Related Issue

CIF-816

## How Has This Been Tested?

manually + added unit test 

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/Utils.java
```diff
@@ -89,6 +93,26 @@ public static Page getSearchResultsPage(Page page) {
...
+    /**
+     * Retrieves the navigation root related to the specified page.
+     * The page and its parents is searched for the navRoot=true property, marking the navigation root.
+     *
+     * @param page the page
+     *
+     * @return the navigation root page if found,  otherwise {@code null}
+     */
+    @Nullable
+    public static Page getNavigationRootPage(Page page) {
```
dplaton: Doesn't `InheritanceValueMap` do the same thing? Try to retrieve a property from up the ancestors?

I know I'm sounding like a broken record, but please do not put stuff in Utils anymore. I know this "utility class" anti-pattern is used throughout AEM, but we're better than this, aren't we? This method belongs in the class that needs it. Whether this method will be needed in the future by other objects, it's not our concern now.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/header/HeaderImpl.java
```diff
@@ -49,4 +53,18 @@ public String getSearchResultsPageUrl() {
 
         return searchResultsPage.getPath()+".html";
     }
+
+    @Override
+    public String getNavigationRootPageUrl() {
+        if (navigationRootPage == null) {
```
dplaton: See above - use `InheritanceValueMap`.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/Utils.java
LSantha: @dplaton , how would you use InheritenceValue for this?

LSantha: @dplaton , @mhaack , @herzog31, I'm not a fan of static utility methods at all but it has a reason why this method ended up here. See the rest of the methods in Utils class and remember who added them. The methods define the site structure and I think we need a better abstraction for that.

dplaton: > how would you use InheritenceValueMap for this?

You wouldn't. You need the page, not the actual value. Sorry for the noise, please disregard.

dplaton: > See the rest of the methods in Utils class and remember who added them.

We all did, I don't deny this. 

> The methods define the site structure and I think we need a better abstraction for that.

One option is to use a Sling Model to which we can adapt the site root page (where "site" is the language root, probably. Not sure how the `@Model` annotation works, I think you can specify a condition so that the adapter is not based on the resource type alone (i.e. adapt it to that particular model when the resource type is `venia/components/structure/page/v1/page` and `navRoot == true`.

LSantha: @dplaton , I would prefer to proceed with this PR further in the scope of the current issue and do the redesign of the code in a dedicated issue and develop a clear solution for it.

cjelger: Honestly I have absolutely no problem with the `Utils` class as long as we don't duplicate code, I'm fine with that.

LSantha: @dplaton , could we clarify everything here or you are still requesting changes?

PR #23, opened 2019-05-15
CIF-851 - AntiSamy warnings in error log when opening a product detail page

herzog31:
## Description
Dynamically loaded HTML from the Magento product description fails the Apache Sling XSS check, implemented via AntiSamy.

* Added sling `XSSAPI` to filter product description HTML as received from Magento.
* Removed purify.js since client-side HTML filtering is no longer needed.

This does not fix the *INFO* messages in the `error.log` file. I discussed with @lkrapf that those log messages are intended behavior of `XSSAPI` and its `AntiSamy` implementation. I just moved the filtering from an implicit Apache Sling filter into our Sling models, to make the filtering process more explicit.

To fix the problem, the following options are available:
* Update the AntiSamy configuration as described in https://helpx.adobe.com/experience-manager/6-5/sites/developing/using/security.html
* Remove unallowed style tags from Magento product descriptions
* Raise log level to not output *INFO*  messages in `error.log` file

## Types of changes

## Checklist:


herzog31:
@LSantha It's not a warning, it's an *INFO* log message and those are completely fine :)
There's nothing in our project we can do directly to remove those messages, as we use the recommended library and this gives us the possibilities I described above. However, I will add the issue and the corresponding fixes to the README, so that customers know what to configure.

PR #28, opened 2019-05-17
CIF-747 - Shopping cart popup

dplaton:
## Description

This PR adds the shopping cart component to the package. The MiniCart is a full client-side component, implemented using plain Javascript. It relies on a `CommerceApi` implementation that provides the connection to the commerce backend.

The shopping cart is created automatically when the page loads and its details are saved in a cookie called `cif.cart`. Once the "Add to cart" action is implemented the cart will be created when the first product is added.

How to test the feature:

1. Install the package and configure accordingly. 
2. Make sure you have your dispatcher instance configured properly and the requests to `http://localhost/magento/rest` and `http://localhost/magento/graphql` are forwarded correctly.
3. Open any page and click the "CartTrigger" icon from the header. The cart should slide from the left and be empty, showing the message "There are no items in your shopping cart"
4. Open your browser's developer console and add a product to the cart by calling the API function:

```javascript
window.CIF.MiniCart.addItem({sku:"VA03-KH-L",qty:"3"})
```
The cart should now display the product (without image, yet) and the cart trigger should display the number of items in the cart.
5. Click on the "kebab" menu next to the cart item and choose "Edit item" - the "edit" panel should slide from the left and allow you to change the quantity in the cart. Clicking on "Update" will close the edit panel and update the cart. 
6. Click on the "kebab" menu next to the cart item and choose "Remove item" - the item should be removed from the cart.

## Related Issue

CIF-747

## Motivation and Context

Provide a client-side reference component for a MiniCart.

## How Has This Been Tested?

Functional testing.

## Types of changes

## Checklist:


ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/CommerceApi.js
```diff
@@ -0,0 +1,204 @@
...
+ *      of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *      Unless required by applicable law or agreed to in writing, software distributed under
+ *      the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ *      OF ANY KIND, either express or implied. See the License for the specific language
+ *      governing permissions and limitations under the License.
+ *
+ *
+ ******************************************************************************/
+'use strict';
```
herzog31: Please move this inside the `(function() { ... })` scope, otherwise it's applied globally to all clientlibs.

ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/CommerceApi.js
```diff
@@ -0,0 +1,204 @@
...
+         */
+        async removeItem(cartQuote, itemId) {
+            return await this._delete(endpoints.guestcarts.itemOperation(cartQuote, itemId));
+        }
+
+
+    }
+
+    function onDocumentReady() {
+        const endpoint = "http://localhost/magento/rest/default/V1";
```
herzog31: This should be `/magento/rest/default/V1` only, so it takes the hostname and protocol from the dispatcher.

ui.apps/src/main/content/jcr_root/apps/venia/components/commerce/minicart/v1/minicart/clientlib/js/templates/minicart-body.js
```diff
@@ -0,0 +1,24 @@
...
+ *     Unless required by applicable law or agreed to in writing, software distributed under
+ *     the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ *     OF ANY KIND, either express or implied. See the License for the specific language
+ *     governing permissions and limitations under the License.
+ *
+ ******************************************************************************/
+
+'use strict';
+
+const minicartBody = `<div class="miniCart__body">
```
herzog31: I would wrap the templates into a `(function() {})()` so the constants are not defined globally in addition to the `window` object.

ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/CommerceApi.js
dplaton: Yes, I'll do that for all the scripts.

ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/CommerceApi.js
dplaton: Right, my bad.

ui.apps/src/main/content/jcr_root/apps/venia/components/commerce/minicart/v1/minicart/clientlib/js/templates/minicart-body.js
dplaton: 👍

ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/CommerceApi.js
```diff
@@ -0,0 +1,203 @@
...
+            if (!response.ok) {
+                let message = await response.text();
+                throw new Error(message);
+            }
+            return response.json();
+        }
+
+        /**
+         * Performs an update operation (POST or PUT).
+         * @param url the request URL
```
cjelger: That parameter is called `endpoint` in the method, and shouldn't this be consistent with `_fetch`? You don't "build" the `url` at the same method level, I think it would be more consistent to do it everywhere at the same place.

ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/CommerceApi.js
```diff
@@ -0,0 +1,203 @@
...
+         * @param sku
+         * @param qty
+         * @param quoteId
+         * @returns {Promise<*>}
+         */
+        async updateCartEntry(cartId, itemId, {sku, qty, quoteId}) {
+            let url = `${endpoints.guestcarts.itemOperation(cartId, itemId)}`;
+            let params = {
+                cartItem: {
+                    sku,
```
cjelger: Silly question but ... shouldn't we have object keys for `sku` and `qty`? Same question for `postCartEntry`

ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/CommerceGraphqlApi.js
```diff
@@ -0,0 +1,94 @@
...
+ *     OF ANY KIND, either express or implied. See the License for the specific language
+ *     governing permissions and limitations under the License.
+ *
+ ******************************************************************************/
+
+
+(function () {
+
+    'use strict';
+    //todo update the proxy rules so that this gets router through the dispatcher
```
cjelger: Should be done in same PR?

ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/CommerceGraphqlApi.js
```diff
@@ -0,0 +1,94 @@
...
+            if (!props.endpoint) {
+                throw new Error('The commerce API is not properly initialized. The "endpoint" property is missing from the initialization object');
+            }
+
+            this.endpoint = props.endpoint;
+        }
+
+        async _fetch(url, params) {
+            let response = await fetch(url, params);
+            if (!response.ok || response.errors) {
```
cjelger: I think we should not throw an error when `response.errors` exist. The GraphQL spec states that a response can contain `data` and `errors` at the same time, and we indeed see this with, for example, the Magento sample data where querying the category tree returns the data and some errors. So as long as there is some `data`, I think it's not fully an error.

ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/CommerceGraphqlApi.js
```diff
@@ -0,0 +1,94 @@
...
+                let message = await response.text();
```
cjelger: Hmm, I think you rather separate the 2 error cases, when not ok, this is fine, but when it's a GraphQL error, then it should be extracted from the errors message, see for example https://github.com/adobe/commerce-cif-connector/blob/master/bundles/cif-connector-graphql/src/test/resources/magento-graphql-error.json

ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/CommerceGraphqlApi.js
```diff
@@ -0,0 +1,94 @@
...
+            return response.json();
+        }
+
+        /**
+         * Retrieves the URL of the images for an array of product data
+         * @param productData a dictionary object with the following structure {productName:productSku}.
+         * The productName is used for filtering the query and the product SKU is used to identify the variant for which to retrieve the image
+         * @returns {Promise<any[]>}
+         */
+        async getProductImageUrls(productData) {
```
cjelger: Do you really need all the images just for the cart? I think it would be simpler to simply take the main thumbnail image, so you also don't need the image url prefix.

ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/PageContext.js
```diff
@@ -0,0 +1,74 @@
...
+        return {
+            cartInfo: cartInfo,
+            maskPage: function () {
+                pageMask.classList.add("mask__root_active");
+            },
+            unmaskPage: function () {
+                pageMask.classList.remove("mask__root_active");
+            },
+            setCartInfoCookie: function ({cartId, cartQuote}) {
+                document.cookie = `cif.cart=${cartId}#${cartQuote}`;
```
cjelger: Use `cookieName` to avoid hard-coding

ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/CommerceApi.js
dplaton: In ES6 { sku } is equivalent to { sku : sku }

ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/CommerceGraphqlApi.js
dplaton: WDYM by "main thumbnail image"?

cjelger: `productData.thumbnail.url`

dplaton: But this is the thumbnail of the product, not the variant, isn't it?

dplaton:
> I honestly think that the amount of stuff you have to do "by hand", like state management, events listeners, css manipulation, handlebars templating, etc, could really be simplified if we use a framework

I think so as well, @cjelger , but this requires a broader discussion. I remember briefly talking about this in the past - if we use a framework for this then we have to use a framework for all our client-side code, otherwise the interaction will still be done through events. 

> Also you haven't added a single unit test yet, and I suppose that it's going to be complicated to test this compared to reusing all the tools available for existing frameworks.

I'm currently doing a research of our options regarding testing the client side components. I'm not talking about UI testing (like Selenium), but real javascript unit-testing, the kind that we have in CIF. Testing tools are universal, they're not tailored for any specific framework. They do have plugins to make testing easier, but at their core they can be used with any Javascript project. Part of this research will also be updating some code to make it more unit-testable (which today it's not).

Based on the latest discussions with Aditi though, I think we'll need to rewrite this with React anyway because when Magento will push the Peregrine components that allows us to easily build a MiniCart we have to use those components (not unlike we'll do with the Search component).

LSantha:
@dplaton , could you please update the issue branch from master? The solution is still using some old dependencies.

dplaton:
@LSantha done.

LSantha:
@dplaton, thank you. Still there's a build failure.

ui.apps/src/main/content/jcr_root/apps/venia/components/structure/header/v1/header/header.html
```diff
@@ -32,6 +32,7 @@
                         d="M1 1h4l2.68 13.39a2 2 0 0 0 2 1.61h9.72a2 2 0 0 0 2-1.61L23 6H6"></path></svg>
                 </span>
                 </button>
+                <sly data-sly-resource="${resource.path}/minicart"/>
```
LSantha: I would suggest to add a test as well here which checks if the minicart child exists because currently AEM almost  gets frozen if minicart is missing. (See the inclusion of navigation above.)

ui.apps/src/main/content/jcr_root/conf/venia/settings/wcm/templates/product-page/structure/.content.xml
```diff
@@ -38,6 +38,8 @@
                     skipNavigationRoot="true"
                     sling:resourceType="venia/components/structure/navigation/v1/navigation"
                     structureDepth="2"/>
+                <minicart jcr:primaryType="nt:unstructured"
```
LSantha: Please make sure to add minicart to all templates. Currently category-page, content-page, landing-page and search-results doesn't have it.

ui.apps/src/main/content/jcr_root/apps/venia/clientlibs/common/js/CommerceGraphqlApi.js
dplaton: @LSantha @cjelger I've updated the proxy configuration to include the URLs to the product images.

PR #31, opened 2019-05-23
CIF-808 - fix sticky Venia header

mhaack:
## Description

Fix the Venia header and make it sticky.

## Related Issue

CIF-808

## Types of changes

## Checklist:

### DIFF
--- a/ui.apps/src/main/content/jcr_root/apps/venia/components/structure/header/v1/header/clientlibs/css/header-component.css
+++ b/ui.apps/src/main/content/jcr_root/apps/venia/components/structure/header/v1/header/clientlibs/css/header-component.css
@@ -44,6 +44,13 @@
     color: rgb(var(--venia-teal));
 }
 
+.header {
+    position: -webkit-sticky;
+    position: sticky;
+    top: 0;
+    z-index: 10;
+}
+
 .header__root {
     background-color: rgb(var(--venia-grey));
     box-shadow: 0 1px rgb(var(--venia-border));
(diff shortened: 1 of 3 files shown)


ui.apps/src/main/content/jcr_root/apps/venia/components/structure/header/v1/header/clientlibs/css/header-component.css
```diff
@@ -44,6 +44,13 @@
...
 }
 
+.header {
+    position: -webkit-sticky;
+    position: sticky;
+    top: 0;
+    z-index: 10;
+}
+
 .header__root {
```
cjelger: I'm not a css expert, but adding replacing `position: fixed` and adding `width: 100%` to `header__root` also fixes that problem

cjelger: It seems "sticky" is useful when you want something to scroll for a while an then at some point stick ... but hey I'm really not a css guru!

LSantha:
@mhaack , in case we need to generate this CSS again from the Venia source code there is a risk that this fix will be lost. This is why I would prefer adding our custom CSS incrementally in a separate file.

mhaack:
@LSantha good point, I moved it to a separate file

ui.apps/src/main/content/jcr_root/apps/venia/components/structure/header/v1/header/clientlibs/css/header-component.css
dplaton: A "sticky" element takes space on the page and the rest of the elements are positioned accordingly. A "fixed" element doesn't take space on the page and it remains in that position. 
However, I see Venia uses "sticky" as well so we might as well go with that.

PR #51, opened 2019-06-13
[CIF-913] Fix shopping cart for non variant products

AKumarSingh:
Fix Shopping cart for non variant products 

## Description

When adding a product to to shopping cart which does not have variant, there is a error in console , and product is not added to cart. This is specific case of a product not having any variant.  

## Related Issue

https://jira.corp.adobe.com/browse/CIF-913 

## How Has This Been Tested?

Adding a product to cart which does not have any variant. 

error screenshot 
Screen Shot 2019-06-13 at 6 56 30 AM

## Types of changes

Added check for variants
Added additional parameter in graphql request for image thumbnail.

## Checklist:


ui.apps/src/main/content/jcr_root/apps/core/cif/clientlibs/common/js/CommerceGraphqlApi.js
```diff
@@ -55,6 +55,9 @@
                     items {
                         sku
                         name
+                        thumbnail {
```
herzog31: To be be consistent with the variants here, we should use `media_gallery_entries` too, instead of `thumbnail`.

mhaack: `thumbnail` has the advantage of providing the full asset url, `media_gallery_entries` only provide relative file paths.

herzog31: Then we can use `thumbnail` for both. However, we should stay consistent. Either one or the other.

mhaack: 👍 for only using one approach

dplaton: 👍 for thumbnail on both, then we can get rid of the prefix concatenation

AKumarSingh: To be clear : I am already using thumbnail here. I am updating the code so that variants also use thumbnail. Right? :)

herzog31: @AKumarSingh yes please :)

PR #62, opened 2019-06-21
CIF-743 - AEM core component - Product Detail with Price (CSR)

herzog31:
## Description

This use case is based one CIF-739. The price listed on the top right corner of the image (attached) needs to be client-side rendered. This is because price could be a personalized product attribute and showing this as client-side rendered in the reference product detail component will provide developers with an example to showcase how server-side rendered content can be enhanced with client-side content.

I added a new function in the `CommerceGraphqlApi.js` file that allows fetching of prices for one or multiple products. The same function can also be used for CIF-804.

## Types of changes

## Checklist:

### DIFF
--- a/ui.apps/src/main/content/jcr_root/apps/core/cif/clientlibs/common/js/CommerceGraphqlApi.js
+++ b/ui.apps/src/main/content/jcr_root/apps/core/cif/clientlibs/common/js/CommerceGraphqlApi.js
@@ -37,6 +37,23 @@
             return response.json();
         }
 
+        async _fetchGraphql(query) {
+            let params = {
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/json'
+                },
+                body: JSON.stringify({ query })
+            };
+
+            let response = await this._fetch(this.endpoint, params);
+            if (response.errors) {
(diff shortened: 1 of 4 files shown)


ui.apps/src/main/content/jcr_root/apps/core/cif/clientlibs/common/js/CommerceGraphqlApi.js
```diff
@@ -37,6 +37,23 @@
...
+            let params = {
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/json'
+                },
+                body: JSON.stringify({ query })
+            };
+
+            let response = await this._fetch(this.endpoint, params);
+            if (response.errors) {
```
dplaton: I remember @cjelger saying something about the `response.errors` not being the best way to check for errors since a response can contain both errors and data.

herzog31: I refactored the code here into a separate function and didn't change existing functionality. I'll take care of that in a follow up PR :)

PR #80, opened 2019-07-03
CIF -879 Improved test cases with better coverage.

AKumarSingh:
## Description

Improved test cases with better coverage.

## Related Issue

CIF-879 

## Motivation and Context

This pull request contains changes in the test case and 

## How Has This Been Tested?

Run test case with build and verify , Verify code coverage. 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes and the overall coverage did not decrease.
- [x] All unit tests pass on CircleCi.
- [x] I ran all tests locally and they pass.

PR #82, opened 2019-07-04
CIF-879 - improve category image loading

mhaack:
## Description

This PR improves the way the category image is loaded. Does not require additional AEM Dispatcher config any more and use the same URL construction (Magento media base url + folder + image) like other components we have.

## Related Issue

CIF-879  

## Motivation and Context

Improve image URLs and align with other components.

## Types of changes

## Checklist:

### DIFF
--- a/dispatcher/conf/dispatcher.any
+++ b/dispatcher/conf/dispatcher.any
@@ -48,9 +48,9 @@
       /rend01
         {
         # Hostname or IP of the render
-        /hostname "aem-hostname"
+        /hostname "mhaack-macbook.local"
         # Port of the render
-        /port "4503"
+        /port "4502"
         # Connect timeout in milliseconds, 0 to wait indefinitely
         # /timeout "0"
         }
--- a/dispatcher/conf/dispatcher.any
+++ b/dispatcher/conf/dispatcher.any
@@ -48,9 +48,9 @@
       /rend01
         {
         # Hostname or IP of the render
-        /hostname "mhaack-macbook.local"
+        /hostname "aem-hostname"
         # Port of the render
-        /port "4502"
+        /port "4503"
         # Connect timeout in milliseconds, 0 to wait indefinitely
         # /timeout "0"
         }
--- a/dispatcher/conf/magento-proxy.conf
+++ b/dispatcher/conf/magento-proxy.conf
@@ -16,7 +16,7 @@ LoadModule proxy_module modules/mod_proxy.so
 LoadModule proxy_connect_module modules/mod_proxy_connect.so
 LoadModule proxy_http_module modules/mod_proxy_http.so
 
-Define magento_host http://my-magento.cloud
+Define magento_host http://checkout-y53b4wa-o24f6fbtz7p34.demo.magentosite.cloud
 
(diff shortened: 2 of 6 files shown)


herzog31:
Feedback

dispatcher/conf/dispatcher.any
```diff
@@ -48,9 +48,9 @@
       /rend01
         {
         # Hostname or IP of the render
-        /hostname "aem-hostname"
+        /hostname "mhaack-macbook.local"
```
dplaton: Replace with  a generic placeholder string

dispatcher/conf/magento-proxy.conf
```diff
@@ -16,13 +16,11 @@ LoadModule proxy_module modules/mod_proxy.so
 LoadModule proxy_connect_module modules/mod_proxy_connect.so
 LoadModule proxy_http_module modules/mod_proxy_http.so
 
-Define magento_host http://my-magento.cloud
+Define magento_host http://checkout-y53b4wa-o24f6fbtz7p34.demo.magentosite.cloud
```
dplaton: Same as above - we should have a placeholder string here

mhaack: Arg, yes I will revert that was from my local testing :-)

dispatcher/conf/dispatcher.any
mhaack: Arg, yes I will revert that was from my local testing :-)

mhaack:
@herzog31 I will wait until https://github.com/adobe/aem-core-cif-components/pull/80 is merged and add more then. Ok?

mhaack:
> Hmm, isn't this an explicit design that we wanted to enforce ALL requests to go via the dispatcher and not directly to the Magento host?

Is that the case? Actually, I think it is the other way around? For the product images which already return absolute URLs we also directly call Magento.

cjelger:
Sorry, forgot about that. But then we should also do the same for the cart images from the cart, so we get all images directly from Magento and remove any image fetching via the dispatcher. Can also be a 2nd PR.

mhaack:
I think for the cart we are ok, it already used https://github.com/adobe/aem-core-cif-components/blob/master/ui.apps/src/main/content/jcr_root/apps/core/cif/clientlibs/common/js/CommerceGraphqlApi.js#L69 to load the product thumbnails with full url.

mhaack:
I think `/magento/img` can be removed from the Apache routing config. But lets check that first and do via second PR.

AKumarSingh:
I have a  small confusion here, Cart is dependent on dispatcher anyway for other things and doesn't work without dispatcher .  so it's a decision of enforcing request via dispatcher or adding a small extra query for getting complete url . So as I understand now, we will not we using image requests via dispatcher in future?

mhaack:
@AKumarSingh the actual cart requests (REST/GraphQL) must be via this patcher because of CORS, for images that is not needed.

> adding a small extra query for getting complete url

This is actually not nice at all but needed since the getCart REST call does not have this information. Once we change the entire cart to GraphQL only this will be optimized.

PR #86, opened 2019-07-09
Refactor Unit Tests to use JavaScript Modules

herzog31:
## Description

* Updated JavaScript component code to export classes. This was they can be easily consumed by other files and especially by our unit tests without exposing anything in the global scope.
* Updated Karma unit tests to import those classes.
* Removed [karma-coverage](https://github.com/karma-runner/karma-coverage) instrumentation because of an outdated version of `istanbul`. It was replaced by `babel-plugin-istanbul` which is executed as part of the webpack build. This blocked the writing of tests for `async/await` constructs in our code.
* Refactored the `webpack.config.js` to allow transpiling of unit tests, while at the same time maintaining the transpiling of the production code.
* Added some unit tests to maintain coverage percentages.

## Types of changes

## Checklist:


ui.apps/package.json
```diff
@@ -8,9 +8,9 @@
     "test": "npm run prettier:check && npm run karma:run",
     "karma:run": "karma start --single-run --browsers ChromeHeadless,FirefoxHeadless karma.conf.js",
     "karma:debug": "karma start --single-run=false --browsers Chrome --debug karma.conf.js",
-    "prettier:check": "prettier --check 'src/**/*.{js,css}' --config ./.prettierrc",
-    "prettier:fix": "prettier --write 'src/**/*.{js,css}' --config ./.prettierrc",
-    "webpack:build": "webpack --config webpack.config.js --mode development"
+    "prettier:check": "prettier --check '{src,test}/**/*.{js,css}' --config ./.prettierrc",
```
dplaton: Is there a reason to have a check and a fix? Can't we just fix the formatting right away?

ui.apps/webpack.config.js
```diff
@@ -26,44 +26,69 @@ const LIB = {
...
-            // Transpile .js files with babel. Babel will by default pick up the browserslist definition in the 
-            // package.json file.
-            {
-                test: /\.js$/,
-                exclude: /(node_modules|bower_components)/,
-                use: {
-                    loader: 'babel-loader',
-                    options: {
-                        presets: ['@babel/preset-env']
+function generaseBaseConfig() {
```
dplaton: It's supposed to be "generate", right? 😁

ui.apps/package.json
herzog31: CircleCi should fail when executing the check, but not do the fix itself, because this would require that it commits the formatted code. So the workflow is:
* Write code
* Check locally with `prettier:fix`
* Commit and push
* CircleCi runs `prettier:check` and fails if not properly formatted

ui.apps/webpack.config.js
herzog31: Fixed :)

PR #88, opened 2019-07-12
CIF-942 - Update client-side components to use GraphQL caching

herzog31:
## Description

* Updated `_fetchGraphql` method to support cached GraphQL requests by using HTTP GET requests and passing the query as query parameter and un-cached requests via HTTP POST.
* Added unit tests for `CommerceGraphqlApi`.
* This will only work with Magento 2.3.2. Magento 2.3.1 does not support GraphQL requests via HTTP GET.
* Refactored `CommerceGraphqlApi` to use JavaScript modules to be easily unit testable.
* Added `@babel/polyfill` to karma, since it'a a runtime dependency after the webpack build.

## Types of changes

## Checklist:

### DIFF
--- a/ui.apps/src/main/content/jcr_root/apps/core/cif/clientlibs/common/js/CommerceGraphqlApi.js
+++ b/ui.apps/src/main/content/jcr_root/apps/core/cif/clientlibs/common/js/CommerceGraphqlApi.js
@@ -33,7 +33,7 @@ class CommerceGraphqlApi {
         return response.json();
     }
 
-    async _fetchGraphql(query, cached) {
+    async _fetchGraphql(query, ignoreCache = false) {
         // Minimize query
         query = query
             .split('\n')
@@ -42,22 +42,22 @@ class CommerceGraphqlApi {
         query = { query };
 
         let params = {
-            method: cached ? 'GET' : 'POST',
+            method: ignoreCache ? 'POST' : 'GET',
             headers: {
                 'Content-Type': 'application/json'
             }
         };
 
         let url = this.endpoint;
-        if (cached) {
+        if (ignoreCache) {
+            // For un-cached POST request, add query to body
+            params.body = JSON.stringify(query);
+        } else {
             // For cached GET request, add query as query parameters
             let queryString = Object.keys(query)
                 .map(k => `${encodeURIComponent(k)}=${encodeURIComponent(query[k])}`)
                 .join('&');
             url += '?' + queryString;
-        } else {
-            // For un-cached POST request, add query to body
-            params.body = JSON.stringify(query);
         }
 
         let response = await this._fetch(url, params);
@@ -103,7 +103,7 @@ class CommerceGraphqlApi {
             }
         }`;
 
-        let response = await this._fetchGraphql(query, true);
+        let response = await this._fetchGraphql(query);
         let items = response.data.products.items;
 
         let productsMedia = {};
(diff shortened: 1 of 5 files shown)


ui.apps/src/main/content/jcr_root/apps/core/cif/clientlibs/common/js/CommerceGraphqlApi.js
```diff
@@ -11,175 +11,183 @@
...
         }
+        return response.json();
+    }
 
-        async _fetch(url, params) {
-            let response = await fetch(url, params);
-            if (!response.ok || response.errors) {
-                let message = await response.text();
-                throw new Error(message);
+    async _fetchGraphql(query, cached) {
```
mhaack: This with later calling it by `_fetchGraphql(query, true);` will always result in a `GET` request. Wouldn't it be worth to make this a config property as well, same like `endpoint`. WDYT?

dplaton: I'm not comfortable with having the request method controlled by a different argument (in this case, `cached`). I'd introduced caching by default (i.e. have an `ignoreCache` attribute) and have a third optional parameter to control the method.

AFAIU we don't get the cached results with POST requests, but I think this is a Magento thing right?

herzog31: It depends on the query if it should be sent in a cache-able query or not. For example use specific queries (like cart) or mutations must be sent as non-cacheable POST query.

For Magento, only GET requests are cached and POST requests are always un-cached. So if we have a global parameter, how do we differentiate between cache-able and non-cache-able data?

dplaton: > that doesn't make sense to me tbh.

No, you're right it doesn't. 

We should still revert the condition, i.e. make the requests always cacheable and ignore the cache as needed. This is how I've seen it working with other clients as well (i.e. Apollo) - everything is cached by default and you have the option to bypass the cache if needed.

mhaack: @dplaton with caching by default (= `GET` by default) we set the minimum Magento version to 2.3.2.
@herzog31 we should add that to the readme/wiki that one can achieve 2.3.1 compatibility by changing to `POST` then.

herzog31: @mhaack So you're referring to an option to globally disable GraphQL `GET` requests to enable 2.3.1 compatibility?

mhaack: Additionally, maybe we should not call it `cached` at all just "request method" or something. Since `GET` requests are cached is Magento behavior and also on Magento side requires additional config, so not all `GET` responses might be cached in the end.

mhaack: > @herzog31 So you're referring to an option to globally disable GraphQL GET requests to enable 2.3.1 compatibility?

Not sure if globally is needed, just document that they are aware of the option.

herzog31: @mhaack I added a small section to the readme.

PR #95, opened 2019-07-25
CIF-992 - Decouple searchbar from header component

herzog31:
## Description

* Created new dedicated searchbar component that displays the search input
* Disabled baseline check to allow for extending the `Header` sling model
* Preparation for moving the header component to the archetype

Also see https://github.com/adobe/aem-cif-project-archetype/pull/20.

## Types of changes

## Checklist:


bundles/core/pom.xml
```diff
@@ -181,14 +181,17 @@
...
                             </instructions>
                         </configuration>
+                        <!-- 
+                        TODO:BASELINE DISABLED UNTIL VERSION 1.0.0
+
                         <executions>
                             <execution>
                                 <id>baseline</id>
                                 <goals>
                                     <goal>baseline</goal>
```
cjelger: I think you can also just set the config option `<skip>true</skip>` (above in the config).

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/header/HeaderImpl.java
```diff
@@ -42,26 +42,19 @@
...
 
-    @Inject
-    private String minicartNodeName;
-
-    private Page searchResultsPage;
     private Page navigationRootPage;
 
     @Override
     public String getSearchResultsPageUrl() {
-        if (searchResultsPage == null) {
```
cjelger: Since you disabled the baseline plugin, why not remove the method from the API if we no longer need it?

parent/pom.xml
```diff
@@ -441,14 +441,17 @@
                     <artifactId>maven-bundle-plugin</artifactId>
                     <version>3.2.0</version>
                     <inherited>true</inherited>
+                    <!--
```
cjelger: Same comment with `skip` config.

herzog31:
@cjelger I added the test.

PR #100, opened 2019-08-21
CIF-758 - Support for multiple category / product pages

cjelger:
## Description

This PR adds the possibility to create "sub-pages" for the generic product and category pages, and configure them to be used for, respectively, a selected set of products or categories.

## Motivation and Context

This makes it easy to add special layouts for some products and categories, without having to make any changes to the navigation items or links to product and category pages.

## How Has This Been Tested?

Added unit tests.

## Types of changes

## Checklist:


cjelger:
To test this, one must first configure the "CIF page filter configuration" in OSGi, for example for the venia demo with the path `/content/venia/.*?/(category-page|product-page)`.

You can then create for example a product page under the main generic product page, go to this page properties, and select some products for which this page will be used. Go to the page editor, add some text to that page to be able to differentiate it, and open the generic product pge for the previously selected product(s). See that the page displays the extra text, although the URL is the one of the generic page.

dplaton:
>To test this, one must first configure the "CIF page filter configuration" in OSGi, for example for the venia demo with the path /content/venia/.*?/(category-page|product-page).

We can create a `cq:OsgiConfig` node in the archetype to avoid going to the configuration console every time

cjelger:
Good point, as I wrote in my previous comment, we should also probably move the 2 new cq-dialog fields to the archetype under the proxy page component.

ui.apps/src/main/content/jcr_root/apps/core/cif/components/renderconditions/pagetemplate/pagetemplate.jsp
```diff
@@ -0,0 +1,48 @@
+<%--
```
dplaton: Is there a reason why this is a JSP and not a servlet? We already have the bulk of the business logic in the `TemplateRenderCondition` class, we might as well make it a servlet.

cjelger: All the documentation and existing components/examples for `granite:rendercondition` were done like that, so I just followed the "guidelines". A JSP is actually a servlet ;-)

dplaton: >A JSP is actually a servlet ;-)

You should talk to some people about this, they'll have a different opinion 😆 

Moving the logic into a bundle protects us from backwards compatibility issues. A JSP is considered to be public API (starting with 6.4 IIRC) so once we release it it's basically locked. I know we are not part of quickstart, but we still have to maintain BC.

mhaack:
+1 of the osgi config in the archetype project, additionally we should all create a second template for both page types
+1 also having 2 cq-dialogs for both page types

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/servlets/SpecificPageFilterConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+package com.adobe.cq.commerce.core.components.internal.servlets;
+
+import org.osgi.service.metatype.annotations.AttributeDefinition;
+import org.osgi.service.metatype.annotations.AttributeType;
+import org.osgi.service.metatype.annotations.ObjectClassDefinition;
+
+@ObjectClassDefinition(name = "CIF page filter configuration")
+public @interface SpecificPageFilterConfiguration {
+
+    // Pattern for venia demo: /content/venia/.*?/(category-page|product-page)
```
mhaack: This should be removed / moved to documentation.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/servlets/SpecificPageFilterFactory.java
```diff
@@ -0,0 +1,103 @@
...
+
+            Object filter = jcrContent.getValueMap().get(SELECTOR_FILTER_PROPERTY);
+            if (filter == null) {
+                continue;
+            }
+
+            // The property is saved as a String when it's a simple selection, or an array when a multi-selection is done
+            String[] selectors = filter.getClass().isArray() ? ((String[]) filter) : ArrayUtils.toArray((String) filter);
+
+            if (ArrayUtils.contains(selectors, selector)) {
```
mhaack: The logic will match and pick always the first child page which matches the selector. In case an author selects one product/category on multiple template pages always the first will be picked. This has to be documented as well.

ui.apps/src/main/content/jcr_root/apps/core/cif/components/structure/page/v1/page/_cq_dialog/.content.xml
```diff
@@ -60,6 +60,33 @@
...
+                                                sling:resourceType="commerce/gui/components/common/cifcategoryfield"
+                                                fieldDescription="Category ids for which this page will be used."
+                                                fieldLabel="Category ids for which this page will be used."
+                                                multiple="{Boolean}true"
+                                                name="./selectorFilter"
+                                                selectionId="id">
+                                                <granite:rendercondition
+                                                    jcr:primaryType="nt:unstructured"
+                                                    sling:resourceType="core/cif/components/renderconditions/pagetemplate"
+                                                    templatePath="/conf/venia/settings/wcm/templates/category-page"/>
```
mhaack: This is (demo) project specific already.

ui.apps/src/main/content/jcr_root/apps/core/cif/components/structure/page/v1/page/_cq_dialog/.content.xml
```diff
@@ -60,6 +60,33 @@
...
+                                                fieldDescription="Product slugs for which this page will be used."
+                                                fieldLabel="Product slugs for which this page will be used."
+                                                multiple="true"
+                                                name="./selectorFilter"
+                                                rootPath="/var/commerce/products"
+                                                selectionId="slug">
+                                                <granite:rendercondition
+                                                    jcr:primaryType="nt:unstructured"
+                                                    sling:resourceType="core/cif/components/renderconditions/pagetemplate"
+                                                    templatePath="/conf/venia/settings/wcm/templates/product-page"/>
```
mhaack: This is (demo) project specific already.

mhaack:
After reviewing the details ... maybe we should have the dialog entirely in the archetype project because https://github.com/adobe/aem-core-cif-components/pull/100/files#diff-719a193b6dd57e0bd8b2ef3808eed1afR74 and https://github.com/adobe/aem-core-cif-components/pull/100/files#diff-719a193b6dd57e0bd8b2ef3808eed1afR88 are already specific to the generated project and will fail for all non `/content/venia/...` projects.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/servlets/SpecificPageFilterFactory.java
cjelger: FYI, picking the first matching child was defined in the JIRA issue, but yes, I will add some documentation
> First generic category and product pages are default

cjelger:
@dplaton @mhaack I moved the `granite:rendercondition` into a servlet and moved the config to https://github.com/adobe/aem-cif-project-archetype/pull/27 - it's ready to be reviewed again.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/servlets/SpecificPageFilterFactory.java
```diff
@@ -0,0 +1,103 @@
...
+import com.day.cq.wcm.api.NameConstants;
+
+@Component(
+    property = {
+        "sling.filter.scope=REQUEST",
+        "sling.filter.methods=GET",
+        "sling.filter.extensions=html"
+    })
+@Designate(ocd = SpecificPageFilterConfiguration.class, factory = true)
+public class SpecificPageFilterFactory implements Filter {
```
mhaack: Just curious why is that actually called `SpecificPageFilterFactory` vs. only `SpecificPageFilter`?

cjelger: Like we did for other components, it's mainly to indicate that we can have multiple instances, each activated by an `OSGi` config. So it's not just one filter, it's configurable and can have multiple instances.

mhaack: Ah right makes sense. I missed `factory = true` param on the annotation.

cjelger:
@mhaack I let you verify and merge the two PRs, and will then update the documentation on the GH wiki.

mhaack:
Verified and created documentation: https://github.com/adobe/aem-core-cif-components/wiki/configuration#multi-catalog-page-template-configuration

PR #114, opened 2019-09-17
CIF-1030 - Testing the React-based components

dplaton:
## Description

Implement the unit testing framework and write some unit tests for the existing React component. These unit tests are not the complete suite, but rather a way to test the capabilities of the framework.

After much debate with myself, I opted for a combination of Jest as a test runner and [React Testing Library](https://testing-library.com/docs/react-testing-library/intro). I chose Jest because it's a natural choice when it comes to testing React components and I chose React Testing Library because it is recommended in the [React testing docs](https://reactjs.org/docs/testing.html). It doesn't rely on DOM selectors (it actually discourages the use of DOM selectors while testing) because that's usually an implementation detail. 

I wrote some snapshot tests as well, although I'm a bit on the fence about how useful are they. The snapshot tests fail as soon as you update the DOM output of a component (you have the option to overwrite the snapshots, of course) so it's easy for someone to mistake an actual failure for a required snapshot update. 

Venia uses React Test Renderer instead of React Testing Library. The difference is that the test renderer works with components directly (i.e. you can look for components directly under the component tree instead of the DOM nodes) but this felt a bit like testing implementation details. 

Please state your feedback in the comments.

### How to run the tests
Run `npm install` to install the dependencies then `npm test` to run the unit tests. In addition to that, you can also run `npm run test:watch` to start Jest in "watch mode" and have it run the tests as you update them.

### CircleCI integration

I also added CircleCI integration using a dedicated job called "jest".

## Related Issue

CIF- 1030

## Motivation and Context

Implement unit testing for the React components

## How Has This Been Tested?
N/A
## Types of changes

## Checklist:


herzog31:
Could we use Karma as test runner as well? Looks pretty similar to me.

dplaton:
Karma spins an actual web browser to run the tests, so this requires a more complicated setup on the React side - you have to build the whole app to run the unit tests. We can definitely look into this option though, because having two test runners is as confusing as React and non-React components in the same project

PR #119, opened 2019-09-30
CIF-1039 - Revisit the MiniCart and Checkout implementation (Part 1)

herzog31:
## Description
* Added state management via `CartProvider` that can be consumed via `useCartState` hook.
    * This object allows to store, manipulate and read the cart state from all child components. It uses a reducer pattern similar to Redux, but native to React.
    * The provider pattern allows us to get rid of passing down all the state and functions to alter the state via props and therefore makes it easier for customers to replace or customise individual components.
    * Additionally to properties in the cart storage, it also includes `addItem` and `removeItem` methods that are used to manipulate the cart directly. These methods get re-created for each cart (e.g. after submitting an order, a new cart is created, also a new set of methods is generated that embed the new cartId).
* Added `CartInitializer` component that takes care of providing an existing cart either from the cookie or a new cart via GraphQL API.
* Updated all components to use `useCartState` hook instead of relying on cart state passed via props.

Error handling as described in the issue is not addressed in this PR, since there are already quite a lot of changes. I will create a follow up PR with error handling and further checkout optimisations.

## How Has This Been Tested?

Tested locally, existing unit tests were adapted.

## Verification
* Test the complete checkout workflow. Make sure that after a complete checkout the cart is reset and you can still add / remove items to / from the cart.
* Delete the cart cookie, don't refresh the page, but open / close the cart. Make sure you can add / remove items to the cart.
* This PR is mostly refactoring, so it should not impact the cart functionality in any way in comparison to before.

## Types of changes

## Checklist:


ui.apps/src/main/javascript/minicart/src/index.js
```diff
@@ -18,15 +18,79 @@ import ApolloClient from 'apollo-boost';
...
+import CartInitializer from './utils/cartInitializer';
 
 const App = () => {
     const client = new ApolloClient({
         uri: '/magento/graphql'
     });
 
+    const [, setCartCookie] = useCookieValue('cif.cart');
+
+    const initialState = {
```
dplaton: This should be moved in a separate module since `index.js` is also a root for the account management components.

ui.apps/src/main/javascript/minicart/src/components/Minicart/minicart.js
```diff
@@ -44,53 +41,39 @@ const MiniCart = props => {
...
     });
+    useEventListener(document, 'aem.cif.add-to-cart', addToCart);
 
-    const removeItemFromCart = useCallback(
-        itemId => {
-            removeItem({ variables: { cartId, itemId } });
-        },
-        [removeItem]
-    );
+    const removeItemFromCart = itemId => {
```
dplaton: This handler should be provided by the context.

ui.apps/src/main/javascript/minicart/src/components/Minicart/container.js
```diff
@@ -11,23 +11,11 @@
...
  *
  ******************************************************************************/
-import React, { useCallback } from 'react';
+import React from 'react';
 import MiniCart from './minicart';
 
-import { useGuestCart } from '../../utils/hooks';
-
 const Container = () => {
-    const [cartId, resetCart] = useGuestCart();
```
dplaton: The sole purpose of the `Container` component was to provide some data to the Minicart. If we're only going to render the `Minicart` component then it doesn't make sense to keep this component around.

ui.apps/src/main/javascript/minicart/src/utils/state.js
```diff
@@ -0,0 +1,29 @@
...
+ *    governing permissions and limitations under the License.
+ *
+ ******************************************************************************/
+
+import React, { createContext, useContext, useReducer } from 'react';
+import { object, func } from 'prop-types';
+
+export const CartContext = createContext();
+
+export const CartProvider = ({ reducer, initialState, children }) => (
```
dplaton: This context provider doesn't contain any cart specific data, so I don't see a point in calling it `CartProvider`. However, I think we should also include the handler functions here so that any component wrapped in this provider can make use of them.

ui.apps/src/main/javascript/minicart/src/components/Minicart/minicart.js
herzog31: I updated the code to make those two methods available via the context. Please also check the update in the PR description.

ui.apps/src/main/javascript/minicart/src/utils/state.js
herzog31: I added the handler functions to `CartInitializer`. Since I added the reducers and default state to this file, I guess we can keep calling it `CartProvider`.

ui.apps/src/main/javascript/minicart/src/components/Minicart/minicart.js
```diff
@@ -26,122 +25,63 @@ import Footer from './footer';
...
-        footer = showFooter ? (
-            <Footer
-                isOpen={isOpen}
-                cart={data.cart}
-                cartId={cartId}
-                handleCloseCart={handleCloseCart}
-                handleResetCart={handleResetCart}
-            />
-        ) : null;
+    // TODO: Ideally this is stored in the cart state
```
dplaton: Yes, ideally the cart state contains a property called `isLoading`.

ui.apps/src/main/javascript/minicart/src/utils/cartInitializer.js
```diff
@@ -0,0 +1,87 @@
...
+    const [{ cartId: stateCartId }, dispatch] = useCartState();
+    const CART_COOKIE = 'cif.cart';
+
+    const [cartId, setCartCookie] = useCookieValue(CART_COOKIE);
+    const [createCart, { data, error }] = useMutation(MUTATION_CREATE_CART);
+    const [addItem] = useMutation(MUTATION_ADD_TO_CART);
+    const [removeItem] = useMutation(MUTATION_REMOVE_ITEM);
+
+    const _createFunctions = (cartId, dispatch) => {
+        return {
```
dplaton: What's the purpose of this function? Couldn't we achieve the same result by using `useCallback`? 
```javascript
removeItem = useCallback( (itemId) => { ... }, 
[cartId, dispatch])
```

herzog31: As discussed in slack, we keep the renamed method to handle two different cases for creating the cart handler functions :)

cjelger:
Sorry @herzog31 , I don't have the capacity to review this PR.

PR #126, opened 2019-10-09
CIF-995 - CIF Core Components customisation pattern

herzog31:
## Description
I created a [customization.md](https://github.com/adobe/aem-core-cif-components/blob/issue/CIF-995/customization.md) file that describes different use cases and how to implement them. Once we agree on the approach as part of this PR discussion, I will move the file to the wiki and link it from the readme.

* Introduction of `Retriever` classes (I'm open to other naming proposals 🐶) which encapsulate the fetching of data with GraphQL queries and decouple it from the Sling models. The same concept is also used to provide placeholder data in the editor context.
* Retrievers are exposed via the `ProductRetriever.java` interface. These can be extended or generalised once we introduce more for more components.
* The `Retriever` allows to set hooks via `.setProductQueryHook(Consumer<U> productQueryHook)`. Hooks are lambdas which can be used to add partial queries to the existing ones. This allows to easily extend the existing queries. Alternatively, a consumer can replace the whole query using the `.setQuery(String query)` method.

I will add follow up pull requests to apply the same concepts to the other components as well, once this one was accepted.

## How Has This Been Tested?
Updated unit tests and manual testing using the given example in a project generated with the archetype.

## Types of changes

## Checklist:


cjelger:
As discussed last week with @herzog31, I have a fundamental issue with the proposed customization pattern. Basically, the `ProductCustomization` interface exports some methods implemented in `ProductImpl`, so it's like a workaround from the fact that `ProductImpl` is an implementation class that we do not OSGi-export. However, having these methods exported via `ProductCustomization` is not any better than directly exporting `ProductImpl`: we are exposing implementation details, and a user would have to understand in details what each of these methods is doing, in order to be able to implement its own customized versions of these methods.

The point I want to make is: we are exposing implementation details anyway, so why not simply OSGi-export `ProductImpl`, declare some methods `protected`, and let people extend `ProductImpl`? I know that from an OSGi and AEM point of view we don't do that, but this is a very valid pattern in object-oriented languages. Now I'm actually being provocative, I don't think we should export `ProductImpl`...

I think maybe a better design would be to move all the data fetching to another class like the `ProductRetriever` that would take care of the GraphQL request, the HTTP call, and would simply return a Magento `ProductInterface`. There would be a setter in the `Product` interface to set the `ProductRetriever`, which would be an interface exposing all the "implementation details" to customize the GraphQL query. This way we remove all GraphQL implementation details from the Sling model so that it better separates the fetching of the data.

I'm still not 100% sure about the exact design, I will try to write a simple example to show you what I mean.

herzog31:
I made some updates on this PR:

* Created `AbstractRetriever` which can be used by any component.
* Created `AbstractProductRetriever` which can be used by components that fetch product data (single product).
* Each component then adds their own retriever based on the two before-mentioned classes. I added two `ProductRetriever` classes to the `Product` and `ProductTeaser` components for now. More components will be added in follow up PRs.

Here are examples of how the `Product` and `ProductTeaser` components can be customized:

```java
package com.venia.cif.core.models;

import com.adobe.cq.commerce.core.components.models.product.Asset;
import com.adobe.cq.commerce.core.components.models.product.Product;
import com.adobe.cq.commerce.core.components.models.product.Variant;
import com.adobe.cq.commerce.core.components.models.product.VariantAttribute;
import com.adobe.cq.commerce.core.components.models.retriever.AbstractProductRetriever;
import com.adobe.cq.commerce.magento.graphql.ProductInterfaceQuery;
import com.shopify.graphql.support.SchemaViolationError;
import org.apache.sling.api.SlingHttpServletRequest;
import org.apache.sling.models.annotations.Model;
import org.apache.sling.models.annotations.Via;
import org.apache.sling.models.annotations.injectorspecific.Self;
import org.apache.sling.models.annotations.via.ResourceSuperType;

import javax.annotation.PostConstruct;
import java.util.List;

@Model(adaptables = SlingHttpServletRequest.class, adapters = MyProduct.class, resourceType = MyProductImpl.RESOURCE_TYPE)
public class MyProductImpl implements MyProduct {

    protected static final String RESOURCE_TYPE = "venia/components/commerce/myproduct";

    @Self @Via(type = ResourceSuperType.class)
    private Product product;

    private AbstractProductRetriever productRetriever;

    @PostConstruct
    public void initModel() {
        productRetriever = product.getProductRetriever();

        // Pass your custom partial query to the ProductRetriever. This class will automatically take care of executing your query as soon
        // as you try to access any product property.
        productRetriever.extendProductQueryWith((ProductInterfaceQuery p) ->
            p.createdAt()
            .addCustomSimpleField("is_returnable"));
    }

    @Override public String getCreatedAt() {
        return productRetriever.getProduct().getCreatedAt();
    }

    @Override public String isReturnable() throws SchemaViolationError {
        return productRetriever.getProduct().getAsString("is_returnable");
    }
    // ...

}
```

```java
package com.venia.cif.core.models;

import com.adobe.cq.commerce.core.components.models.productteaser.ProductTeaser;
import com.adobe.cq.commerce.core.components.models.retriever.AbstractProductRetriever;
import com.adobe.cq.commerce.magento.graphql.ProductInterfaceQuery;
import org.apache.sling.api.SlingHttpServletRequest;
import org.apache.sling.models.annotations.Model;
import org.apache.sling.models.annotations.Via;
import org.apache.sling.models.annotations.injectorspecific.Self;
import org.apache.sling.models.annotations.via.ResourceSuperType;

import javax.annotation.PostConstruct;

@Model(adaptables = SlingHttpServletRequest.class, adapters = MyProductTeaser.class, resourceType = MyProductTeaserImpl.RESOURCE_TYPE)
public class MyProductTeaserImpl implements MyProductTeaser {

    protected static final String RESOURCE_TYPE = "venia/components/commerce/myproductteaser";

    @Self @Via(type = ResourceSuperType.class)
    private ProductTeaser teaser;

    private AbstractProductRetriever productRetriever;

    @PostConstruct
    public void initModel() {
        productRetriever = teaser.getProductRetriever();

        // Pass your custom partial query to the ProductRetriever. This class will automatically take care of executing your query as soon
        // as you try to access any product property.
        productRetriever.extendProductQueryWith((ProductInterfaceQuery p) ->
            p.createdAt());
    }

    @Override public String getCreatedAt() {
        return productRetriever.getProduct().getCreatedAt();
    }

    // ...
}

```

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/retriever/AbstractProductRetriever.java
```diff
@@ -0,0 +1,216 @@
...
+        }
+        return this.product;
+    }
+
+    /**
+     * Stores a product.
+     *
+     * @param product Product
+     */
+    protected void setProduct(ProductInterface product) {
```
cjelger: Similar comment as earlier regarding the use of `protected` fields vs. `protected` methods. It's a matter of taste so feel free to decide.

bundles/core/src/test/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/ProductRetrieverTest.java
```diff
@@ -0,0 +1,91 @@
...
+        String sampleQuery = "{ my_sample_query }";
+        retriever.setQuery(sampleQuery);
+        retriever.getProduct();
+
+        verify(mockClient, times(1)).execute(sampleQuery);
+    }
+
+    @Test
+    public void testExtendedProductQuery() {
+        retriever.extendProductQueryWith((ProductInterfaceQuery p) -> p.createdAt()
```
cjelger: Related to my previous comment with having `ProductInterfaceQueryDefinition` here as the method argument: this would remove the cast, and would prevent the possibility to use any other query type here.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productteaser/ProductTeaserImpl.java
```diff
@@ -81,99 +69,79 @@ protected void initModel() {
...
         }
         return null;
     }
 
     @Override
     public String getUrl() {
-        if (product != null) {
-            return SiteNavigation.toProductUrl(productPage.getPath(), baseProductSlug, variantSku);
+        if (getProduct() != null) {
+            return SiteNavigation.toProductUrl(productPage.getPath(), productRetriever.getProduct().getUrlKey(), combinedSku.getRight());
```
cjelger: Why don't you use `getProduct()` here like in other methods?

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productteaser/ProductTeaserImpl.java
```diff
@@ -186,29 +154,15 @@ private SimpleProduct findVariant(ConfigurableProduct configurableProduct, Strin
...
-            .name()
-            .image(i -> i.url())
-            .price(generatePriceQuery());
+    private NumberFormat getPriceFormatter() {
+        if (priceFormatter == null) {
+            // Initialize NumberFormatter with locale from current page.
+            // Alternatively, the locale can potentially be retrieved via
+            // the storeConfig query introduced with Magento 2.3.1
+            Locale locale = currentPage.getLanguage(false);
+            priceFormatter = Utils.buildPriceFormatter(locale, productRetriever.getProduct() != null ? getCurrency() : null);
```
cjelger: Why don't you use getProduct() here like in other methods?

customization.md
```diff
@@ -0,0 +1,200 @@
+# DRAFT: Customizing CIF Core Components
```
cjelger: Is this still a draft or shall we also review that?

herzog31: This can be reviewed as well. I added `DRAFT` to indicate that this file will not be merged, but moved into the wiki.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productteaser/ProductTeaserImpl.java
herzog31: The previous implementation was explicitly using the base product to get the slug. `getProduct()` could return a variant if the combined sku includes a variant sku.

herzog31:
@cjelger I added your feedback 👍  Please check again.

customization.md
```diff
@@ -0,0 +1,200 @@
...
+
+This document describes customization of CIF Core Components on the layers of Sling models and data retrieval via GraphQL. For all other customizations, please follow the patters described in [Customizing Core Components](https://docs.adobe.com/content/help/en/experience-manager-core-components/using/developing/customizing.html).
+
+
+## Use Cases
```
mhaack: I would change this to "Customization Case" or "Customization Scenario"

customization.md
```diff
@@ -0,0 +1,200 @@
...
+## Use Case to Layer Mapping
+
+| Use Case | Template | Interface | Implementation | Data Layer | Schema |
+| -------- | -------- | --------- | -------------- | ---------- | ------ |
+| A        | ✔️       | ✔️        | 🔶            | ✔️          | ✔️    |
+| B        | 🔶       | ❌        | 🔶            | 🔶          | ✔️    |
+| C        | 🔶       | ❌        | 🔶            | ❌          | ❌    |
+| D        | ❌       | ❌        | ❌            | ❌          | ❌    |
+
+✔️ no adaption needed, ❌ adaption needed, 🔶 partial re-use possible
```
mhaack: User other icon then red X, red X looks like not possible on the first look

mhaack:
Verified with Wiki sample

PR #133, opened 2019-10-24
CIF-1062 Implement the account creation process

dplaton:
## Description

This feature will allow visitors to create an account. The back-end interaction is handled by the `UserContext` and the user input is handled by the `CreateAccount` component.

Steps for verification:
1. Go to the Sign In form
2. Click the "Create Account" button
3. Fill the form with the required information - first name, last name, email, password, password confirmation
4. Click `Submit`

The user is immediately logged in after the account is created and the "My Account" section is shown. Check that the information matches the one entered in the account creation form.

Sign out and repeat the process. Check that the "My account" information is consistent with the new information entered in the "Create Account" form.

## Related Issue
CIF-1062
## How Has This Been Tested?
Functional testing and unit testing
## Types of changes

## Checklist:


ui.apps/src/main/javascript/minicart/src/components/CreateAccount/createAccount.js
```diff
@@ -0,0 +1,124 @@
...
+    const [{ createAccountError, isSignedIn, isCreatingCustomer }, { createAccount }] = useCreateAccount();
+
+    const handleCreateAccount = formValues => {
+        createAccount(formValues);
+    };
+
+    const errorMessage = createAccountError ? createAccountError : null;
+
+    if (isSignedIn) {
+        showMyAccount();
```
herzog31: I would add a `return` after this line to make sure the evaluation of this component is stopped, since you switch to another view anyway.

ui.apps/src/main/javascript/minicart/src/context/UserContext.js
```diff
@@ -57,16 +64,45 @@ const UserContextProvider = props => {
...
+        onError: error => {
+            let createAccountError = parseError(error);
+            setUserState({ ...userState, createAccountError });
+        }
+    });
+
+    /*
+     * We use the apollo client directly
+     */
+    const getCustomerDetails = opts => {
```
herzog31: Can you elaborate why you have to use the Apollo client directly here and can't use the hook you used before?

dplaton: I already added a comment (actually, I forgot to write the whole comment above 🙈) to explain this. I stumbled into an issue with the following flow: 
- create an account
- sign out
- create another account - the "My Account" panel showed the data of the first account. 

The reason was that the data returned by the first `getCustomerDetails` call was still available so that the code just returned that. This whole "hook-a-palooza" needs to be refactored a bit, maybe even using the client directly or just refactor the components to issue their own query.

herzog31: Got it, thanks :) Looks good for now, let's improve it later.

PR #131, opened 2019-10-21
CIF-1081 - Client side product price loading does not respect multi store hierarchy configuration

herzog31:
## Description

* Added `StoreViewExporter` model to expose Magento store view in DOM.
* Updated `CommerceGraphqlAPI` library to pass the `Store` header accordingly.

## How Has This Been Tested?

* Updated unit tests.

## Verification
* Open a product or category page and verify that the client-side GraphQL request that retrieves pricing information contains a `Store: default` header.
* Configure the website to use a different storeview, e.g. `venia`. You can do that by setting the `cq:magentoStore` property on a page. Verify that the client-side pricing requests now send a `Store: venia` header.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes and the overall coverage did not decrease.
- [X] All unit tests pass on CircleCi.
- [X] I ran all tests locally and they pass.

PR #138, opened 2019-11-01
CIF-1058 - Cart/Checkout do not respect multi store hierarchy configuration

herzog31:
## Description

Similar to https://github.com/adobe/aem-core-cif-components/pull/131 the React components that use the Apollo client to send GraphQL requests now respect the `cq:magentoStore` property to include a `Store` header in the requests.

## Verification
1. Add a product to the cart. Verify that the client-side GraphQL requests (both for adding the product and retrieving the cart) contain a `Store: default` header.
2. Configure the website to use a different storeview, e.g. `venia`. You can do that by setting the `cq:magentoStore` property on a page. Repeat step 1 and verify that the header has changed.

## Types of changes

## Checklist:

### DIFF
--- a/ui.apps/src/main/javascript/minicart/src/index.js
+++ b/ui.apps/src/main/javascript/minicart/src/index.js
@@ -25,12 +25,11 @@ import { CheckoutProvider, initialState as initialCheckoutState, reducer } from 
 import AuthBar from './components/AuthBar';
 
 const App = () => {
-
     const storeView = document.querySelector('body').dataset.storeView || 'default';
 
     const client = new ApolloClient({
         uri: '/magento/graphql',
-        headers: { 'Store': storeView }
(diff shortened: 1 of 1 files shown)


ui.apps/src/main/javascript/minicart/src/index.js
```diff
@@ -25,8 +25,11 @@ import { CheckoutProvider, initialState as initialCheckoutState, reducer } from
 import AuthBar from './components/AuthBar';
 
 const App = () => {
+    const storeView = document.querySelector('body').dataset.storeView || 'default';
```
dplaton: Ideally, this would be passed directly into the props via that web component magic that we used in the past.
We'll leave it like that for now, but we have to address that pretty soon as we introduce component configurations.

PR #139, opened 2019-11-06
CIF-1073 - Reorganize the React components project

dplaton:
## Description

Move the React components folder directly under the parent folder of the CIF Core Components project. 
Update the project information - description, license etc.
Put all the React components scripts into a single client library.

To be merged with https://github.com/adobe/aem-cif-project-archetype/pull/55

## Related Issue

CIF-1110

## Motivation and Context

Move the React project code from the CIF Code Components content package.

## How Has This Been Tested?

Unit tests and functional tests

## Types of changes

## Checklist:


react-components/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+-   Minicart
+-   Checkout
+-   Sign In
+-   My Account
+-   Password recovery (Forgot Password)
+
+## Building the project
+
+The project is built using the command `npm run build`. The build process bundles all the code into one client library which is placed in `../ui.apps/src/main/content/jcr_root/apps/core/cif/clientlibs/react-components/dist`.
```
cjelger: I have two main concerns with that entire change, but maybe I missed something:
- it seems this project is not integrated with the main "reactor" build, but it must also be built, otherwise the clientlib is going to be missing in ui.apps, right?
- I then find it a bit "tricky" to build something and add it to another project (ui.apps), because it basically creates a dependency that isn't reflected now at the project level ... can we somehow fix that? At least to make sure that the reactor project builds everything in the right order?

dplaton: The `react-components` project is built by the `ui.apps` project, see https://github.com/adobe/aem-core-cif-components/pull/139/files#diff-d6ca0a35a41bb1932eddf863e9f17311R199

cjelger: ah ok, sorry, I thought you had also removed that and really wanted the two sub-projects to be independent.

dplaton:
@herzog31 The more I think of it the more I realized that we need a more complex solution here, along the lines that you mentioned - have the "Demo" components project that has the `react-components` project as a dependency (and maybe just does the rendering)

herzog31:
Maybe. Think about the customisation use case:
* The main logic of the react components lives in the `react-components` project.
* To make the react components work ootb with our components, they need to be used by `ui.apps`. So ideally we have a minimal clientlib that just imports everything.
* To customize the components, customers would overlay this clientlib in the archetype. They can import our react components from npm, adding their code and generating a new clientlib that overlays the old one.

@dplaton wdyt?

dplaton:
@herzog31 They can either overlay the clientlib or the `customfooterlibs.html` script of the page component to include their own client libraries (could be more than one).

But yes, that's the pattern I have in mind - create a "demo" project which just imports the components and renders them.

As a side note, this is *exactly* what the Venia team is doing now. I wonder if it wouldn't make more sense to channel our efforts into making their components render in AEM (as I remembered we talked during the workshop)

herzog31:
@dplaton Do we need a dependency of `ui.apps` to the react project? I think otherwise the build order is not guaranteed.

Also for the client-side AEM components, we had prettier configured. Could you please add this to the react components as well (including `npm test`)?

dplaton:
Yes, we need a (Maven) dependency for that purpose, although I'm now trying to see how this would work for SNAPSHOTs (i.e. `npm link`).

react-components/package.json
```diff
@@ -1,19 +1,25 @@
 {
-  "name": "minicart",
+  "name": "@adobe/aem-core-cif-react-components",
```
herzog31: Can we keep the version of the `@adobe/aem-core-cif-react-components` npm project in sync with the maven reactor project version?

cjelger: Good point, I think this could be done in the POM by adding a an extra step `npm-module-prepare` that would do `npm version ${project.version}`, and would be bound to the right phase/goal during `release:prepare`. I think you could use the `completionGoals` option of the `maven-release-plugin` to do that at the right time.

dplaton: @cjelger @herzog31, how "blocking" is this issue for us now? I've been trying to find a suitable solution for a few days now and nothing seems to work properly. Maven's `completionGoals` runs only Maven goals and doesn't run `scm:add` and `scm:commit` properly. I even created my own `release.js` script to handle the version update and Git commit, but that is not executed correctly by the `exec-maven-plugin` (can't even see the output from it).

For this release, we can sync the version manually (i.e. just put `0.6.0` in the version field) and continue the work after this Friday's release, what do you think?

cjelger: Sure no problem, we can fix it later, it's not important for this release.

herzog31: 👍

dplaton: I added a script to automatically update the version of the package and commit the changes. This script can be ran manually via `npm run update-version -- --version=x.x.x`. I tried to convince Maven to run it but I couldn't get the right version, it would always call it with the `SNAPSHOT`

PR #146, opened 2019-11-11
CIF-1123 - FeaturedCategoryList component defines an absolute resourc…

cjelger:
…e path

- removed absolute path
- upgrade surefire test plugin to latest version

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -380,7 +380,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


parent/pom.xml
```diff
@@ -380,7 +380,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
-                    <version>2.18.1</version>
+                    <version>3.0.0-M3</version>
```
mhaack: How is that relevant for this fix?

cjelger: Not relevant at all, I just thought I use the same PR to also update that.

PR #147, opened 2019-11-11
CIF-1108 - Update WCM core components dependency to 2.7.0

herzog31:
See https://github.com/adobe/aem-cif-project-archetype/pull/54.


mhaack:
Verified manually with sample project built from issue/CIF-1108 branch.

PR #150, opened 2019-11-22
CIF-1042 - Support Braintree payment option in checkout

herzog31:
## Description

* Added support for Braintree credit card and PayPal payment methods.
* Braintree drop in will appear below the payment method selection.
* Forms for payment method, billing address and shipping method are now only accessible after providing a shipping address.
* There's a limitation currently, that only the payment method and a payment nonce are stored in the cart. After a website refresh only the payment method is available, but no details like PayPal email or last digits of the credit card. When you go back to the payment form, the Braintree form will be empty and all information have to be provided again.

I added some documentation in the wiki: https://github.com/adobe/aem-core-cif-components/wiki/Braintree-Payment-Provider.

## Screenshot
Imgur

## How Has This Been Tested?

* Unit tests
* Manual testing using https://developers.braintreepayments.com/reference/general/testing/node (can also be used for verification)

## Types of changes

## Checklist:

### DIFF
--- a/ui.apps/src/main/javascript/minicart/src/components/Checkout/editableForm.js
+++ b/ui.apps/src/main/javascript/minicart/src/components/Checkout/editableForm.js
@@ -119,7 +119,6 @@ const EditableForm = props => {
 
             // Store payment method in state, so we can later use it in the
             // payment step of the checkout.
-            // TODO: Store payment method description as well via hidden field?
             dispatch({ type: 'setPaymentMethod', paymentMethod: args.paymentMethod });
 
             switch (args.paymentMethod.code) {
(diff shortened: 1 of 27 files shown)


ui.apps/src/main/javascript/minicart/src/components/Checkout/editableForm.js
```diff
@@ -96,7 +115,37 @@ const EditableForm = props => {
...
+                    setBraintreePaymentMethodOnCart({
+                        variables: {
+                            cartId: cartId,
+                            paymentMethodCode: args.paymentMethod.code,
+                            nonce: args.paymentNonce
+                        }
+                    });
+                    break;
+                }
+                case 'braintree_paypal': {
```
cjelger: You can merge this `case` with the first one, the code seems exactly the same.

herzog31: I fixed that 👍

herzog31: Thanks, I fixed that 👍

PR #152, opened 2019-11-28
CIF-995 - CIF Core Components customisation pattern (Part 3)

herzog31:
See https://github.com/adobe/aem-core-cif-components/pull/126.

Changes applied for `ProductCarousel` component.


mhaack:
Verified

PR #157, opened 2019-12-06
CIF-915 - Create JS unit tests for Navigation component

LSantha:
* added unit test for navigation component
 * improved clicking experience on down navigation arrow

## Related Issue

CIF-915

## How Has This Been Tested?

Manually, with unit tests.

## Types of changes

## Checklist:


LSantha:
Ping.

ui.apps/test/components/commerce/navigation/navigationTest.js
```diff
@@ -0,0 +1,312 @@
...
+        };
+        body = window.document.querySelector('body');
+        navigationRoot = document.createElement('div');
+        body.insertAdjacentElement('afterbegin', navigationRoot);
+        navigationRoot.insertAdjacentHTML(
+            'afterbegin',
+            `
+                <div class="header__primaryActions">
+                    <button class="navTrigger__root clickable__root" aria-label="Toggle navigation panel">
+                        <span class="icon__root">
```
dplaton: This is only a "test" HTML so I think you can trim it down a little bit. One example - you don't need the full SVG here. As a general rule, we only need the elements defined by a selector so we can simulate clicks.

ui.apps/test/components/commerce/navigation/navigationTest.js
```diff
@@ -36,30 +36,23 @@ describe('Navigation', () => {
                 <div class="header__primaryActions">
                     <button class="navTrigger__root clickable__root" aria-label="Toggle navigation panel">
                         <span class="icon__root">
-                            <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
-                                <line x1="3" y1="12" x2="21" y2="12"></line>
-                                <line x1="3" y1="6" x2="21" y2="6"></line>
-                                <line x1="3" y1="18" x2="21" y2="18"></line>
+                            <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
```
dplaton: >you don't need the full SVG here

What I meant to say is that you don't need the SVG *at all* here.

LSantha: Actually I need some svg at places, because I click on the svg in the test.

dplaton: But the actual event is not handled by the SVG, but by the `button` element, at least this is how it should be.

LSantha: Right, the event handler is registered on the button.

LSantha: @dplaton , do you need any more clarification?

PR #158, opened 2019-12-10
CIF-1115 - Allow product dnd on product carousel

cjelger:
- added sling POST processor to add products to the array property
- modified carousel to support products paths and skus

## Description

This PR adds the possibility to add products to the carousel with drag'n'drop, and extends the array of products with each new product added to the carousel.

## How Has This Been Tested?

Added unit tests and manually tested.

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/servlets/MultiFieldDropTargetPostProcessor.java
```diff
@@ -0,0 +1,155 @@
...
+ *      ...
+ *      propertyName="./dropTarget->@books">
+ *   </books>
+ * </cq:dropTargets>
+ * }
+ * </pre>
+ * 
+ */
+@Component(service = { SlingPostProcessor.class })
+public class MultiFieldDropTargetPostProcessor implements SlingPostProcessor {
```
herzog31: Would it possible to scope this servlet somehow to our CIF components? I'm wondering what would happen if a customer already has a custom drop target servlet like this for other components?

At least we should mention this in the documentation or component readme.

cjelger: It's kind of scoped because it only handles property keys starting with `./dropTarget->` so it could only "collide with itself" if a customer already uses the same post processor. I will extend the documentation or component readme.

PR #164, opened 2019-12-12
CIF-1153 - Generic related product component

cjelger:
- created new component based on product carousel
- added unit tests
- adapted carousel component and sling model to support the new component

## How Has This Been Tested?

Added unit tests, and manually tested in AEM.

Screenshot 2019-12-13 at 09 55 35

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/relatedproducts/RelatedProductsRetriever.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/relatedproducts/RelatedProductsRetriever.java
@@ -31,7 +31,19 @@ import com.adobe.cq.commerce.magento.graphql.QueryQuery;
 class RelatedProductsRetriever extends AbstractProductsRetriever {
 
     static enum RelationType {
-        RELATED_PRODUCTS, UPSELL_PRODUCTS, CROSS_SELL_PRODUCTS
+        RELATED_PRODUCTS("Related products"),
+        UPSELL_PRODUCTS("Upsell products"),
+        CROSS_SELL_PRODUCTS("Cross-sell products");
+
+        private final String text;
+
+        private RelationType(String text) {
+            this.text = text;
+        }
+
+        String getText() {
+            return text;
+        }
     }
 
     static enum ProductIdType {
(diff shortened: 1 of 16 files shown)


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/relatedproducts/RelatedProductsRetriever.java
```diff
@@ -0,0 +1,117 @@
...
+import com.adobe.cq.commerce.magento.graphql.ProductInterface;
+import com.adobe.cq.commerce.magento.graphql.ProductInterfaceQueryDefinition;
+import com.adobe.cq.commerce.magento.graphql.ProductPricesQueryDefinition;
+import com.adobe.cq.commerce.magento.graphql.ProductsQueryDefinition;
+import com.adobe.cq.commerce.magento.graphql.QueryQuery;
+
+class RelatedProductsRetriever extends AbstractProductsRetriever {
+
+    static enum RelationType {
+        RELATED_PRODUCTS, UPSELL_PRODUCTS, CROSS_SELL_PRODUCTS
```
mhaack: If we design the enum with key & name we could use a more author-friendly name for the dialog data source.

cjelger: You're right, I'll change this to display user-friendly names.

dplaton: > If we design the enum with key & name we could use a more author-friendly name for the dialog data source.

...and be able to translate those names as well.

cjelger: I added human-readable text, and they should be translatable with the use of `ResourceBundle`.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/productcarousel/ProductCarousel.java
```diff
@@ -27,7 +27,7 @@
 public interface ProductCarousel {
```
mhaack: One more thing. I'm thinking we should rename interface now as this now more than just a carousel thing. But I'm not really creative with a more common name.

cjelger: Well it's still a carousel, why change that? Only the source of the data changes ... but it's still  a carousel of products.

mhaack: Well was just an idea :-) lets keep it.

PR #179, opened 2020-01-15
CIF-1045 - [Research] Edit custom PDP / PLP page

cjelger:
## Description

This PR is a possible solution to fix the editing issue of "specific pages": the idea as detailed by @mhaack in CIF-1045 is to disable the `SpecificPageFilter` on author (actually except when `WCMMode.DISABLED` is set) and also have product and category URLs point to the specific pages. This way, on author one will see and be able to edit specific pages and links, while on publish all these "deep links and specific pages" will be hidden.

## How Has This Been Tested?

Extended unit tests and manually verified everything on publish.

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/utils/SiteNavigation.java
```diff
@@ -166,4 +189,81 @@ public static String toProductUrl(String pagePath, String slug, String variantSk
...
+    /**
+     * Builds and returns a product page URL based on the given page, slug, and variant sku.
+     * This method might return the URL of a specific subpage configured for that particular page.
+     * 
+     * @param page The page used to build the URL.
+     * @param slug The slug of the product.
+     * @param variantSku An optional sku of the variant that will be "selected" on the product page, can be null.
+     * @return The product page URL.
+     */
+    public String toProductUrl(Page page, String slug, String variantSku) {
```
dplaton: If `SiteNavigation` has the request object then all the methods asking for the `Page` should be refactored to just use the request path from the request object.

cjelger: That would be too easy, links are not created for and to the current page ;-) For example, on the category page you create links to product pages.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/utils/SiteNavigation.java
```diff
@@ -14,24 +14,33 @@
...
 
 import com.day.cq.commons.inherit.HierarchyNodeInheritanceValueMap;
 import com.day.cq.commons.inherit.InheritanceValueMap;
+import com.day.cq.commons.jcr.JcrConstants;
+import com.day.cq.wcm.api.NameConstants;
 import com.day.cq.wcm.api.Page;
 import com.day.cq.wcm.api.PageManager;
+import com.day.cq.wcm.api.WCMMode;
 
 public class SiteNavigation {
```
mhaack: Not 100% related to this research but I think we should have this URL generation an OSGI service. So customers can bring their own implementation to generate PDP / PLP URLs based on their needs?

cjelger: Yes it's a good idea, customers will surely want to customise that.

PR #182, opened 2020-01-23
CIF-1169 - [React Components] Use maven-frontend-plugin for building the project

### LINKED ISSUE #156: Build requires npm to be installed
The maven build of the reactor breaks if no npm binary is installed. This could be avoided if the maven-frontend-plugin is used, which allows to freely specify which version of node and npm should be used.
It also makes it easier to build the project without maintaining these dependencies manually.

cjelger:
- replace exec plugin with frontend plugin in react-components and ui.apps
- automate the upgrade of the NPM release version of the react-components project
- change react-components project packaging to 'pom' because it doesn't generate any jar

## Motivation and Context

This ensures the NPM builds do not rely on locally installed versions of npm and node so the build is consistent and reproducible.

Fixes #156.

## How Has This Been Tested?

The `release:prepare` was tested locally: it modifies all the POM files and also the `package.json` and `package-lock.json` files in one single commit as expected (see screenshot).

Screenshot 2020-01-23 at 10 10 19

## Checklist:


cjelger:
Fixes #156

PR #199, opened 2020-02-13
CIF-1188 - Use default implementations in Sling models consistently

cjelger:
- removed all "default" methods in sling model interfaces


mhaack:
Verified

PR #213, opened 2020-02-26
CIF-1166 - Internationalization support for AEM client-side Core Components

herzog31:
## Description

* Use [i18next](https://www.i18next.com/) library for internationalization in React components.
* Added mock i18next configuration for React jest testing. This uses the actual language files.
* Added `jsonTransform.js` and configuration in Jest which allows to directly import JSON files in tests, like `import cart from '../i18n/en/cart.json';`.
* Added `i18next-scanner` tool which automatically extracts translated strings from the components. Use `npm run translation:scan` to generate language files.
* Replaced the Magento Peregrin `Price` component with a custom one that supports `i18next`. So now the prices are displayed according to the `i18n` configuration.
* The `i18n` library is configured to take the language from the `<html lang="en">` tag, from the path or from the subdomain (in this order) and download language files accordingly (via AJAX from the clientlib).
* Language files are available to the React components from the `react-components` clientlib as per recommendation from the refsquad team.
* Added translations for the `Minicart` component.
* Since the language files are downloaded asynchronously, it is recommended to use the React [Suspense](https://reactjs.org/docs/concurrent-mode-suspense.html) component to show a loading screen. Once this is supported by the Apollo GraphQL client, it could be used for it as well.
* Building the `react-components` clientlib in `ui.apps` now also copies the language files from the `react-components` project which are generated by `i18next-scanner`.

Follow up work:
* Add translations for `MyAccount` and `Checkout` components.
* `react-components` clientlib in `ui.apps` should be moved to the archetype since we can expect customization to happen. In terms of i18n, we can expect customers to customize the way `i18next` downloads the language files (maybe from an external service).

## How Has This Been Tested?

* Unit test
* Manual testing with second language

## Types of changes

## Checklist:


ui.apps/src/main/javascript/react-components/copy-translation.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ ******************************************************************************/
+
+const fs = require('fs');
+const path = require('path');
+
+const languageFolder = './node_modules/@adobe/aem-core-cif-react-components/i18n/en';
+const clientlibFolder = '../../content/jcr_root/apps/core/cif/clientlibs/react-components/resources/lang/en-US';
+
+fs.mkdirSync(clientlibFolder, { recursive: true });
```
cjelger: Shouldn't we rather do that copy in some npm script in `package.json` or maybe better, in the POM of the `ui.apps` project? The `maven-resources-plugin` has a `copy-resources` goal, so we would avoid that the react-components project has to know "where" it is consumed.

herzog31: Done 👍

PR #221, opened 2020-03-10
CIF-1182 - Use the same property type in components with product DnD

cjelger:
- make it possible to configure the property type to be saved when drag'n'dropping a product

## How Has This Been Tested?

TO BE DONE: extend unit tests

## Types of changes

## Checklist:


ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/productteaser/v1/productteaser/_cq_editConfig.xml
```diff
@@ -7,6 +7,11 @@
             jcr:primaryType="cq:DropTargetConfig"
             accept="[.*]"
             groups="[product]"
-            propertyName="./selection"/>
+            propertyName="./multiDropTarget->@selection">
```
mhaack: Why needs the product teaser the "multiDropTarget"?

cjelger: Because we implement that DnD "save" feature in the `MultiFieldDropTargetPostProcessor` so we don't have to introduce another post-processor. I added the `multiple` parameter to control the ADD or REPLACE behaviour, so the teaser uses `false` so DnD'ing another product replaces the selection.

mhaack: +1 for keeping everything in on post-processor. But this looks cumbersome somehow because it is not "AEM standard".

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/servlets/MultiFieldDropTargetPostProcessor.java
```diff
@@ -119,6 +137,20 @@ private void processProperty(Resource resource, String target, String propertyVa
             resourceResolver.delete(dropTargetResource);
         }
 
+        if (selectionType != null) {
+            if (SKU.equals(selectionType) || COMBINED_SKU.equals(selectionType)) {
+                propertyValue = StringUtils.substringAfterLast(propertyValue, "/");
+            } else if (SLUG.equals(selectionType)) {
+                Resource product = resourceResolver.getResource(propertyValue);
```
mhaack: The SLUG resolving part must be changed, this only works because of the virtual product bundles are installed by the connector. We must not have any dependency on virtual products within the components project. This is will also very likely break with the new authoring tools.

So maybe two simple reasons to move the entire post-processor to the connector project. That is IMHO the better place since it does something which is related to the authoring tools. Then this code part would also be ok.

cjelger: I agree it's something more related to authoring so we can move it in the connector project. But strictly speaking, this doesn't really remove the dependency, because for example the `carousel` DnD target would still be defined with `./multiDropTarget->@product`, so it wouldn't work without the connector.

Note that we could also simply remove the "slug" option: both the carousel and the teaser anyway do not support having product(s) configured by slug, so I only added it to demonstrate it's possible to support that, but we don't have any use case for this.

cjelger:
As agreed with the team, the post-processor was moved from this repository to the connector repo, see https://github.com/adobe/commerce-cif-connector/pull/105

ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/productteaser/v1/productteaser/_cq_editConfig.xml
mhaack: I wonder if we can find a way to get both property flavors: `propertyName="./selection"` and  `propertyName="./multiDropTarget->@product"` working?

The first format is the AEM "standard" and ideal for components which do only accept d&d of one product this should be used.

cjelger: Well if we want to allow the "standard" DnD **AND** the choice of `selectionId` (sku, path, slug), we must have a way to "detect" that special DnD feature. A possible alternative to what is being done here could be:
```
propertyName = "./selection"
<parameters
  selectionId="multiDropTargetValue->sku@selection">
```
We need 2 things:
- we must "indicate" that this POST must be processed by the `MultiFieldDropTargetPostProcessor` (so we need a special `String` somewhere to "activate" the post processor)
- we have to indicate which property will hold the value. This is required because the DnD POST sets all properties in the same HTTP form, so we have to make sure we know which property must be modified before saving.

So I think it's possible, but would add more complexity to the post-processor, and anyway requires a special configuration in `cq:editConfig` ... so I don't really see a big benefit. If someone only needs the standard existing DnD, they can just continue to use this.

mhaack: Ok, I think I can agree, let's keep the current approach then.

mhaack:
Verified

Issue #198, opened 2020-02-12
Feature Request: Allow dynamic filtering of search results


kevin-mitchell:
### User Story
As a user I want to able to use the filtering or "faceting" functionality provided by Magento to more easily narrow down search results.

### Description & Motivation

Magento allows for dynamic filtering of search results based on attribute configuration within Magento. It would be nice if the search related components supported some of this functionality. The "dynamic" part of this comes in because the attributes that are configured for search within Magento are dynamic and so too the AEM components used should also be dynamic so that as the searchable attributes are updated in Magento the attributes too are updated within AEM.

### Deliverables

* Additional components
* Modification to existing search, navigation bar, and potentially product list UI elements
* Ability to make simple introspection queries is required to be able to dynamically query Magento's GraphQL endpoints

### Acceptance Criteria

_After setting up CIF and adding the appropriate search component(s) to a page..._

* Search displays with a list of available filters
* Adding a new attribute as filterable in Magento should result in the option being available for filtering in AEM
* Removing an attribute from Magento should also remove the attribute as a filter option from the AEM search steps

mhaack:
Fixed by #223

mhaack:
Thanks for contributing @kevin-mitchell

PR #223, opened 2020-03-11
#198 Feature/add graphql dynamic filters introspection

kevin-mitchell:
New feature added allowing for dynamic filtering (or "faceting") of attributes on search and product list pages.

## Description

This PR represents the work required to provide filtering functionality on the search and product list pages, powered by Magento 2.3.4's new GraphQL filtering functionality.

There are quite a few changes here a number of key items to call out.

### SearchResultsService

The `SearchResultsService` is intended to be useful way of querying for Magento products that acts as the translation layer between "frontend" code / Sling Models / etc and the GraphQL query layer.  The hopeful goal of this class is to make it usable in a number of potential different cases - it's original use was in the `SearchResults` component, however it was easily incorporated into `ProductList` as well.

### DRY-ness

These changes were originally made specifically for search results, NOT for the category pages (aka `ProductList` and related). Later, we needed to support the filtering on the category page as well. As mentioned above, the flexibility of the `SearchResultsService` and the fact that the category page is VERY similar to the search results made this fairly simple, HOWEVER the result is that there is quite a bit of code that is nearly identical between the two components `SearchResults` and `ProductList`. 

There are enough differences between the two designs that I didn't want to totally steamroll over the `ProductList` component but it would likely be possible to refactor to reduce some of the code duplication at some point.

Certain aspects of the category page vs search are unique, e.g. category specific images and other marketing related features made possible by the differences in `ProductList`.

### Dependency on [GraphQL Introspection additions](https://github.com/adobe/commerce-cif-magento-graphql/pull/13)

The PR "[#91 add css rules for search results component relating to filter bar](https://github.com/adobe/commerce-cif-magento-graphql/pull/13)" has changes that are REQUIRED for this PR to land. Most likely we'll need to bump version numbers somewhere in here to make sure that this change lands in such a way that the updated GraphQL introspection changes are available. 

## Related Issue

#198 

## Motivation and Context

Motivated by Magento 2.3.4 and it's ability to dynamically filter search results.

## How Has This Been Tested?

There are a large number of unit tests that have been added as part of this change. Where possible I've used the [converter pattern](https://java-design-patterns.com/patterns/converter/) when changing GraphQL layer data to Sling Model friendly POJOs and the like. This makes testing quite a bit easier.

Screen Shot 2020-03-11 at 8 37 49 PM

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImpl.java
```diff
@@ -15,10 +15,10 @@
 package com.adobe.cq.commerce.core.components.internal.models.v1.productlist;
```
kevin-mitchell: The changes made to this class are almost entirely in an effort to port the functionality from the `SearchResults` component here, so that essentially the same functionality available in search is available on the product listing pages (basically the category pages)

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java
```diff
@@ -29,7 +29,9 @@
 import com.adobe.cq.commerce.graphql.client.RequestOptions;
```
kevin-mitchell: Changes to this file are made so that introspection queries can be made. If memory serves originally I had created a separate `MagentoIntrospectionGraphqlClient` or something similar but the team thought it would be better to just have the existing class handle both queries types (with the default behavior remaining in tact)

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/searchresults/SearchResultsImpl.java
```diff
@@ -11,14 +11,12 @@
  *    governing permissions and limitations under the License.
```
kevin-mitchell: This was refactored to use the `SearchResultsService`, which hopefully hides most of the details about how to actually query for search results.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/productlist/ProductList.java
```diff
@@ -71,21 +71,15 @@
     @Nullable
     String getTitle();
 
-    int getTotalCount();
```
kevin-mitchell: Here again an attempt was made to DRY things out a bit by moving all pagination / search result information and pagination functionality into the search result class(es).

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/searchresults/SearchResults.java
```diff
@@ -11,22 +11,62 @@
...
+/**
+ * Don't forget the comment
+ */
+@ProviderType
 public interface SearchResults {
 
+    /**
+     * Name of the String resource property indicating number of products to render on front-end.
+     */
+    String PN_PAGE_SIZE = "pageSize";
```
kevin-mitchell: Some of this functionality was borrowed from the `ProductList` component to allow the same authoring options (at least those that are relevant for search result pages)

ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/searchresults/v1/searchresults/paginationbar.html
```diff
@@ -0,0 +1,101 @@
...
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
+<template data-sly-template.pagination="${@ searchResultsSet}">
+	<div class="category__pagination">
+		<div class="category__pagination_root">
+			<sly data-sly-list.page="${searchResultsSet.paginationParameters}">
+				<!--/*First Page button & Previous page button inactive */-->
+				<sly data-sly-test="${pageList.first && searchResultsSet.searchOptions.currentPage==1} ">
+					<a class=" category__pagination_navbutton-inactive category__pagination_navbutton-arrow category__pagination_tilebutton" aria-label="move to the first page">
+                        <span class="category__pagination_icon-root">
+                            <svg xmlns="http://www.w3.org/2000/svg" width="1.2rem" height="1.2rem" viewBox="0 0 24 24" fill="#999" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><polygon points="11 19 2 12 11 5 11 19"></polygon><polygon points="22 19 13 12 22 5 22 19"></polygon></svg>
```
dplaton: We should extract these in CSS

ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/productlist/v1/productlist/paginationbar.html
```diff
@@ -14,10 +14,15 @@
 <template data-sly-template.pagination="${@ productList}">
```
LSantha: The pagination bar needs to reflect the changes in made in commit 7496ebf4007081559711f31d9adf7979a94ffda9. What I see here seems to be based on an earlier version.
This applies to the pagination bar of the search results too.

kevin-mitchell: Done! I actually pulled all of the pagination related code into a new `Pager` class that is a LOT easier to unit test now in isolation and without a bunch of mocking. Hopefully this works for you!

ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/searchresults/v1/searchresults/paginationbar.html
kevin-mitchell: I think this makes a LOT of sense and would be ideal, but this pattern seems to be used in a number of places in the repo so I hesitate to add yet another new convention in this already massive PR. That said, is there a place this has already been accomplished that you're aware of (replacing the svg with a CSS rule for these particular elements?)

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/CategoryRetriever.java
```diff
@@ -17,48 +17,13 @@
...
-import com.adobe.cq.commerce.magento.graphql.ProductInterfaceQueryDefinition;
-import com.adobe.cq.commerce.magento.graphql.ProductPriceQueryDefinition;
 
 class CategoryRetriever extends AbstractCategoryRetriever {
 
     CategoryRetriever(MagentoGraphqlClient client) {
         super(client);
     }
 
-    private ProductPriceQueryDefinition generatePriceQuery() {
```
kevin-mitchell, line 30: This class no longer is responsible for generating product searches, this is all being handled in the search service

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/retriever/AbstractCategoryRetriever.java
```diff
@@ -129,26 +123,6 @@ public void extendCategoryQueryWith(Consumer<CategoryTreeQuery> categoryQueryHoo
...
-     * {@code
-     * categoryRetriever.extendProductQueryWith(p -> p
-     *     .createdAt()
-     *     .addCustomSimpleField("is_returnable"));
-     * }
-     * </pre>
-     *
-     * @param productQueryHook Lambda that extends the product query
-     */
-    public void extendProductQueryWith(Consumer<ProductInterfaceQuery> productQueryHook) {
```
kevin-mitchell, line 148: I moved this same functionality into the SearchService, and the extended product query hook is supplied as a parameter in the search method.

bundles/core/src/test/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImplTest.java
```diff
@@ -142,183 +153,54 @@ public void getImageWhenMissingInResponse() {
...
-                Assert.assertNull(item.getImageURL());
-            } else {
-                Assert.assertEquals(smallImage.getUrl(), item.getImageURL());
-            }
-
-            Assert.assertEquals(productInterface instanceof GroupedProduct, item.getPriceRange().isStartPrice());
-        }
-    }
-
-    @Test
```
kevin-mitchell, line 187: These tests (or similar tests) have been moved to the `Pager` class which is responsible for computing pagination.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImpl.java
```diff
@@ -104,27 +106,30 @@ private void initModel() {
...
         }
 
-        locale = productPage.getLanguage(false);
-
         MagentoGraphqlClient magentoGraphqlClient = MagentoGraphqlClient.create(resource);
 
         // Parse category id from URL
-        final String categoryId = parseCategoryId();
+        categoryId = parseCategoryId(this.request.getRequestPathInfo().getSelectorString(), request.getParameter(
+            "category_id"));
```
cjelger, line 125: Why do you also check the `category_id` parameter from the URL? I saw that the filters on the category pages include that `category_id` request parameter, why do we need this? (since we anyway get it from the url selector)

I'm asking because we will have a merge conflict here with https://github.com/adobe/aem-core-cif-components/pull/248 so I'd like to fully understand if this is needed or not.

bundles/core/src/main/java/com/adobe/cq/commerce/core/search/internal/models/SearchOptionsImpl.java
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Processes the search term to prepare it for the actual query.
+     * This method just prepends/appends the default wildcard character % to the search term. Overriding methods can implement their own
+     * processing.
+     *
+     * @return the processed search term, by default {@code "%searchTerm%"}
+     */
+    @Nonnull
+    private String processSearchTerm(String searchTerm) {
```
cjelger, line 114: This seems to be obsolete, you no longer add the % characters, right? Can we remove that?

kevin-mitchell: Originally this was done as it gave "wildcard" matches back to the search results, but you're right, the comment at least is obsolete because from what I can tell from the current Magento documentation:

> Wildcards are no longer supported in products queries.

I can / will remove this method, but I guess we should just be aware that in that case there is no filtering done on the search query (which I suppose is probably best, allow Magento to reject / deal with invalid query terms)

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImpl.java
kevin-mitchell: It was quite a while ago I wrote this and I'm not sure if at the time I added this as a fallback so the query string parameters alone could control all of the options, but that said I can remove this.

That said, as it is currently in `master` there is a null check for this value but not for integer value / etc so this was partially why I refactored to return an `Optional`. For now I will revert to the version you have in `master` though with the understanding this is improved / updated in the PR you referenced!

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImpl.java
```diff
@@ -202,122 +186,56 @@ public boolean loadClientPrice() {
...
-        if ((getTotalCount() % navPageSize) == 0) {
-            navPagesSize = getTotalCount() / navPageSize;
-            navPageNext = (navPageCursor < (getTotalCount() / navPageSize)) ? (navPageCursor + 1) : navPageCursor;
-        } else {
-            navPagesSize = (getTotalCount() / navPageSize) + 1;
-            navPageNext = (navPageCursor < ((getTotalCount() / navPageSize) + 1)) ? (navPageCursor + 1) : navPageCursor;
+    @Override
+    public SearchResultsSet getSearchResultsSet() {
+        if (searchResultsSet == null) {
+            searchResultsSet = searchResultsService.performSearch(searchOptions, resource, productPage, request);
```
herzog31, line 224: Setting a query hook for the product query is supported via the 5th argument of the `performSearch` search method. However, currently a customer has no way of using it in any way using the Sling model delegation pattern. Replacing the whole query is also not possible.

The same applies for the `SearchResults` component.

cjelger: Right I missed that, I had this in my radar and then forgot to check it ... We definitely have to support this.

PR #234, opened 2020-03-22
CIF-1072 - Support Registered Cart+Checkout

dplaton:
## Description

In order to support registered cart and checkout, some changes had to be made in the implementation of the `UserContext`. The `UserContext` only exposes the user state now, leaving the data fetching responsibility to the components that need it. One example is the `SignIn` component which now does its own data fetching instead of calling the UserContext for that. This made the cart merging operations (essential for this story) easier to implement.

The `CartInitializer` component has been rewritten to take the user state into account as well. 

I only kept the necessary `useEffect` hooks and switch to the async / await for mutations. 

The shipping / billing address is not auto-completed. This will be part of a follow-up issue.

Note:
I left the `console.log` statements in place for debugging. I will remove them once the verification is done, prior to merging.

## Related Issue

CIF-1072

## Motivation and Context

Support the Registered Shopper story

## How Has This Been Tested?

Functional tests, unit tests

## Types of changes

## Checklist:


react-components/src/components/App/app.js
```diff
@@ -20,24 +20,33 @@ import ApolloClient from 'apollo-boost';
...
 const App = props => {
     const { uri, storeView = 'default' } = props;
 
     const client = new ApolloClient({
         uri,
-        headers: { Store: storeView }
+        headers: { Store: storeView },
+        request: operation => {
+            let token = checkCookie('cif.userToken') ? cookieValue('cif.userToken') : '';
+            operation.setContext({
```
herzog31: Ideally we should only set the header if a token is set. Otherwise we end up with an invalid `'Authorization: Bearer '` header.

react-components/src/components/AuthBar/authBar.js
```diff
@@ -22,8 +22,13 @@ import { func } from 'prop-types';
...
 const AuthBar = props => {
     const { showSignIn, showMyAccount } = props;
+    const [{ currentUser, isSignedIn }, { getUserDetails }] = useUserContext();
 
-    const [{ currentUser, isSignedIn }] = useUserContext();
+    useEffect(() => {
+        if (isSignedIn && currentUser.email === '') {
+            getUserDetails();
+        }
+    }, [getUserDetails]);
```
herzog31: Is this `useEffect` section executed again if you logout and log in again? It looks like the `getUserDetails` function never changes, so this section is only executed once.

react-components/src/components/Checkout/receipt.js
```diff
@@ -38,11 +31,14 @@ const Receipt = () => {
...
                     {/* prettier-ignore */}
                     <Trans i18nKey="checkout:order-confirmation">
-                        The order number is {{ orderId : order.order_id }}. You will receive an order confirmation email with order
+                        The order number is {{ orderId : orderId }}. You will receive an order confirmation email with order
                         status and other details.
                     </Trans>
                 </div>
-                <Trigger action={continueShopping}>
+                <Trigger
+                    action={async () => {
```
herzog31: Why don't you call `continueShopping()` directly? The `async` function you pass is asynchrone the same way like `continueShopping()`.

react-components/src/components/Checkout/useOverview.js
```diff
@@ -0,0 +1,46 @@
...
+    const [{ cart, cartId }, cartDispatch] = useCartState();
+
+    const [placeOrder] = useMutation(MUTATION_PLACE_ORDER);
+    const [inProgress, setInProgress] = useState(false);
+
+    const submitOrder = async () => {
+        setInProgress(true);
+        try {
+            const { data } = await placeOrder({ variables: { cartId } });
+            setInProgress(false);
```
herzog31: You can move the `setInProgress(false)` in a `finally` block.

react-components/src/components/AuthBar/authBar.js
dplaton: Yes, it is executed again. The dependency function changes every time the `fetchUserDetails` function changes.

react-components/src/components/App/app.js
dplaton: Ack

react-components/src/components/Checkout/receipt.js
dplaton: No reason, I think it's a leftover from previous tests. I'll call it directly

PR #241, opened 2020-03-27
CIF-1317 - Support Virtual Products

herzog31:
## Description

* Add support for virtual products in Product and ProductTeaser components.
    * Use `addVirtualProductsToCart` mutation to add virtual products to cart.
* Update cart to hide shipping address and method for fully virtual carts (carts that only contain virtual products).
* Fix bug where the street of the billing address was not properly stored.

## How Has This Been Tested?

* Updated unit tests
* Manual test with registered and guest user

## Types of changes

## Checklist:


react-components/src/queries/query_cart_details.graphql
```diff
@@ -1,5 +1,6 @@
 query cartDetails($cartId: String!) {
     cart(cart_id: $cartId) {
+        is_virtual
```
cjelger: Is this `true` only when all the cart products are virtual?

react-components/src/components/Minicart/cartInitializer.js
```diff
@@ -55,7 +57,10 @@ const CartInitializer = props => {
                 });
                 dispatch({ type: 'open' });
                 dispatch({ type: 'beginLoading' });
-                return addItem({
+
+                let addItemFunc = ev.detail.virtual ? addVirtualItem : addItem;
```
cjelger: Hmmm, shouldn't we have that `virtual` attribute per item added to the cart? What if we have a grouped product composed of two virtual products and one normal product? Not sure if this is realistic, but it would be better to support that, WDYT?

react-components/src/queries/query_cart_details.graphql
herzog31: Yes. As soon as one physical item is in the cart, this property is `false`.

react-components/src/components/Minicart/cartInitializer.js
herzog31: I will try if a grouped product can be composed of virtual and physical products.

We would need to split the query then however into `addSimpleProductsToCart` and `addVirtualProductsToCart` mutations.

dplaton: Not sure if this works OOTB in Magento, but I've seen countless of "Guitar, amp and 1yr subscription to online course" offers, so it's a valid business case.

herzog31: I created a follow up JIRA for this and will work on it after the release :)

PR #246, opened 2020-04-01
CIF-1337 - Support virtual products that are part of grouped products

herzog31:
## Description

This PR is on top of https://github.com/adobe/aem-core-cif-components/pull/241. Will change the base branch to master after the before-mentioned PR was merged.

* Restored original event format for add to cart events. If a product is virtual is now defined via a property on each item of the event.
* Added separate mutation calls for physical and virtual products. Both might be called at the same time for grouped products that contain physical or virtual products.
* Updated tests.
* Minor refactoring in the addToCart component code.

## How Has This Been Tested?

* Updated unit tests

## Types of changes

## Checklist:


bundles/core/src/test/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/ProductImplTest.java
```diff
@@ -360,6 +360,12 @@ public void testGroupedProduct() throws IOException {
             Assert.assertEquals(pi.getName(), item.getName());
             Assert.assertEquals(pi.getPriceRange().getMinimumPrice().getFinalPrice().getValue(), item.getPriceRange().getFinalPrice(), 0);
             Assert.assertEquals(gp.getItems().get(i).getQty(), item.getDefaultQuantity(), 0);
+
+            if (i < 3) {
```
cjelger, line 364: To avoid the check based on the position, you could do 
`Assert.assertEquals(pi.get("__typename").equals("VirtualProduct"), item.isVirtualProduct());`

react-components/src/components/Minicart/cartInitializer.js
```diff
@@ -47,24 +47,44 @@ const CartInitializer = props => {
...
+                    promises.push(
+                        addVirtualItem({
+                            variables: { cartId, cartItems: virtualCartItems },
+                            refetchQueries: [{ query: CART_DETAILS_QUERY, variables: { cartId } }],
+                            awaitRefetchQueries: true
+                        })
+                    );
+                }
+
+                return Promise.all(promises)
```
cjelger, line 87: I have two comments here:
1. Why don't we do the two mutations at the same time, I mean in the same GraphQL request?
2. Whether we do 1. or not, and like with `Promise.all`, don't we have here a typical case of possible "lost update" because of the concurrent calls? Depending on the ordering and/or concurrency, it could well be that the last CART_DETAILS_QUERY would miss one add-to-cart and would only show the result of one request.

I don't know if there are any guidelines here in the Magento documentation to better understand the possible race condition.

bundles/core/src/test/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/ProductImplTest.java
cjelger: Or `Assert.assertEquals(pi instanceof VirtualProduct, item.isVirtualProduct());`

react-components/src/components/Minicart/cartInitializer.js
herzog31: I discussed with @dplaton and there are basically two ways of approaching this problem. We can define a third query which combines both mutations like

```graphql
mutation addBundleToCart( params ) {
  addVirtualProductsToCart( params ) {
    cart {
      id
    }
  }
  addSimpleProductsToCart( params ) {
    cart {
      id
    }
  }
}
```

and use it in case we have both simple and virtual products. This might become problematic, if we need to support additional product types (e.g. bundled products?). Alternatively, we could let Apollo automatically batch requests https://blog.apollographql.com/batching-client-graphql-queries-a685f5bcd41b, but this solution is actually not recommended.

I couldn't observe a race condition as of now, since there seem to be some internal optimisations in Apollo that make sure the cart refetch is done only once after both mutations. But I can adapt the code to make sure.

cjelger: It's up to you, if you checked that the refetch is only done once at the right time, then it's fine. I thought there would maybe be an option in the Apollo client to "merge" multiple GraphQL mutations in one single request.

cjelger: The batching link has this part:

> This isn’t suggesting to write one large GraphQL query at the container-level. Instead, write queries normally, next to the components that use them. When you’re ready to optimize a section of an app, convert those queries to fragments and export them from the component file. You can then import these fragments in the container, let the container make the single, large query, and pass the fragment results back to the children.

So maybe instead of writing all possible combinations, we could use fragments and merge them as needed?

herzog31: I combined both queries now which also makes sure that the refetching is done only once. So no more race conditions.

PR #248, opened 2020-04-02
CIF-1256 - Better customizable PLP/PDP URL generation

cjelger:
Related PR: https://github.com/adobe/aem-cif-project-archetype/pull/100

## Description

This PR adds an OSGi service to provide configurable PDP and PLP urls. See the related PR in the archetype repo for an example. The documentation will be added to the wiki after the PR is merged.

The PR is large but the main code to review is in `UrlProvider` and `UrlProviderImpl`. The rest is the refactoring to replace `SiteNavigation` by `UrlProvider` in all the components and tests.

## How Has This Been Tested?

## Types of changes

## Checklist:


bundles/core/pom.xml
```diff
@@ -89,6 +89,10 @@
                             javax.annotation;version=0.0.0,
                             *
                         </Import-Package>
+                        <Embed-Dependency>
+                            commons-text;inline=false;scope=compile|runtime
```
mhaack, line 93: Instead of embedding here I suggest we add as a dependency of ui.apps package and install commons-text directly. The jar is already an OSGI bundle.

cjelger: Done, the bundle is now embedded in `ui.apps`.

PR #249, opened 2020-04-02
CIF-1322 - Cache GraphQL response for Navigation component

LSantha:
* added caching for categories in Navigation component
 * updated unit tests

## Description

The navigation component makes a GraphQL request which returns the same results for most of the time. This request is sent for every page rendering which makes it the most frequent GraphQL request, this makes it a candidate for caching.

## Related Issue

CIF-1322

## How Has This Been Tested?

Manually and JUnit.

## Types of changes

## Checklist:


cjelger:
Actually we wouldn't even need the new method in `MagentoGraphqlClient` because we have access to the current `Resource`, so we use the component resource type to use the right cache. This would be nicer so we don't have to modify the components code, it would be "transparent" for the components if the response comes from the cache or not.

mhaack:
+1 for @cjelger suggestion, if we can get this solved in a more transparent and generic way. 

I was also thinking we could implement this via the underlying HTTP client and use `CachingHttpClientBuilder` which will respect the cache information of the HTTP response (which should be available for Magento if via Fastly or Varnish) or with a custom cache config.

LSantha:
@cjelger , I was also thinking about the store view and finally dropped it because as far as I remember if you specify a different store view you will have a different root category id (and sub category id) too.

The other issue with the storeview is that currently it's picked up from the context aware configuration by the GraphQL client transparently to the components. Having to expose the storeview to the Navigation component for the sake of caching looks a like a code a design issues to me that we would need to avoid. One way of doing that is to make the connection object suitable to be part of a cache key (add meaningful equals() and hashCode() method based on the connection parameters including store view). Then if you make any modification in the context aware configuration could affect the cached data, that will automatically handled by the cache on the next access instead of cache expiration. This would allow ad-hoc component level caches like the one in this PR. 

Obviously the other alternative is to push the cache down from component level to lower layers like to the connection or extra layer like we have it in the connector.

herzog31:
> I was also thinking about the store view and finally dropped it because as far as I remember if you specify a different store view you will have a different root category id (and sub category id) too.

This is wrong. If you translate certain categories, they will have the same ids across storeviews.

LSantha:
@mhaack , @cjelger I also agree that there are other solutions to this problem including more generic solutions like in the connector. I just would like to remind you gust that this PR and the direction it's taking is the results of our team discussion last week where we decided together to address in this manner the performance issue with the Navigation component affecting the rendering of all pages.

LSantha:
@herzog31 , is there a catalog where I could check that? I can fix it if needed.

herzog31:
@LSantha I don't have any at hand. But you can follow this guide to set this up locally: https://magefan.com/blog/translate-category-in-magento-2.

cjelger:
> I just would like to remind you guys that this PR and the direction it's taking is the results of our team discussion last week where we decided together to address in this manner the performance issue with the Navigation component affecting the rendering of all pages.

@LSantha In the last meeting @mhaack presented his findings with SDI, and then you said at the end of the meeting that you had another solution and you'll open a PR. We didn't discuss your proposal and didn't decide on the technical solution. So no we didn't "decided together to address in this manner".

floriniordache:
@LSantha Please integrate all incoming feedback, I'm sure everyone agrees that an open PR with the actual code changes that all can see and review have precendence over any other principial discussion. We're all looking for a good solution, so everyone's input is important.

mhaack:
@LSantha we should avoid point solutions and since you already agreed on a more generic approach ...
> I also agree that there are other solutions to this problem including more generic solutions

... let's go that route. So I reject this PR.

There are a few more generic options, which are also "more standard". Maybe for example if we use HTTP caching headers and they would work immediately for other areas as well. Maybe even with less code.

LSantha:
@cjelger, I explained my solution earlier, and after @mhaack retracted his proposal, you said that then go with my proposal. Then @mhaack agreed and added the conclusion to the issue as a comment. You can go and check it.

cjelger:
Just as a note: HTTP caching would only work with GET requests (but the client can be configured to use POST), and we'd have to make sure it doesn't cache GraphQL `mutation` requests, and I'm not sure how it handles different HTTP headers (= store-view). So this would need extra investigation, and would only work if the server returns proper HTTP caching directives.

floriniordache:
> @cjelger, I explained my solution earlier, and after @mhaack retracted his proposal, you said that then go with my proposal. Then @mhaack agreed and added the conclusion to the issue as a comment. You can go and check it.

@LSantha I thought everything is up to discussion up to, and after, merging the PR, no? Don't we want the proposal to be the best one we can get, according to everyone's feedback?

mhaack:
> Just as a note: HTTP caching would only work with GET requests (but the client can be configured to use POST), and we'd have to make sure it doesn't cache GraphQL mutation requests, and I'm not sure how it handles different HTTP headers (= store-view). So this would need extra investigation, and would only work if the server returns proper HTTP caching directives.

Only for GET is fine, mutations via POST will not be cached. But you are right is requires config on the Magento side to return caching directive header fields.

dplaton:
>  we'd have to make sure it doesn't cache GraphQL `mutation` requests, 

Do we use GraphQL mutations anywhere?

LSantha:
I've fixed the store view, just in case we decide to use this PR.

PR #252, opened 2020-04-09
CIF-1331 - The React cart component should not create a cart by default

dplaton:
## Description

The old implementations created cart handlers specific for a cart id. This was a good implementation, but it doesn't fit the world in which we want the anonymous cart to be created when the first item is added. Since we don't have a `cartId` from the beginning I had to make some changes around how the cart operations are handled - all those operations had been moved to a separate `actions` module and not bound by a specific cart id.

## Related Issue

CIF-1331

## Motivation and Context

Prevent the creation of anonymous carts with every visit to the page

## How Has This Been Tested?

Functional testing, unit tests.

## Types of changes

## Checklist:


react-components/src/actions/cart/actions.js
```diff
@@ -0,0 +1,169 @@
...
+ * @param {Object} payload a parameters object with the following structure:
+ *      cartDetailsQuery - the query object to execute
+ *      createCartMutation - the mutation to execute to create the cart
+ *      addToCartMutation - the mutation to execute to add the item to the cart
+ *      dispatch - the dispatch callback for the cart context
+ *      cartItems - the items to add to the cart
+ *      cartId - the id of the cart in which to add the items. This could be `undefined` or `null`
+ */
+
+export const addItemToCart = async payload => {
```
herzog31, line 28: In the `addItemToCart` action you take care of the loading dispatches, while for `removeItemFromCart`, `removeCoupon` and `addCoupon`, this is done in the `cartInitializer`. Can we somehow harmonise this?

dplaton: You're right, we should avoid dispatching UI state changes from the actions themselves. I moved the dispatch call to `useMinicart`

PR #255, opened 2020-04-20
CIF-1329 - Reset the context data between checking out and showing the receipt

herzog31:
## Description

* Reset the cart automatically after submitting a new order. For anonymous users this will not automatically create a new cart, but only after they add an item to the cart.
* Checkout state is only reset after a user clicks on "Continue shopping" or reloads the page. Otherwise order information would not be available to be displayed on the order confirmation page.

## How Has This Been Tested?

* Added unit tests

## Types of changes

## Checklist:


react-components/src/context/UserContext.js
```diff
@@ -144,6 +146,14 @@ const UserContextProvider = props => {
         }
     };
 
+    const resetCustomerCart = async () => {
```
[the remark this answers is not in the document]
herzog31: I refactored the function into a separate action file.

react-components/src/context/UserContext.js
```diff
@@ -144,6 +147,10 @@ const UserContextProvider = props => {
...
+        await resetCustomerCartAction({ fetchCustomerCartQuery, dispatch });
```
dplaton, line 151: Not sure if we should proxy this call. The components can bring their own query (in case of customizations) so maybe we should get that as an argument.

herzog31: I added the query as a parameter to the exposed function.

herzog31:
@dplaton can you have a look again please?

PR #256, opened 2020-04-22
CIF-1321 - Empty Commerce tab in page properties

LSantha:
Added commerce tab in page properties with appropriate render conditions for custom PDPs and PLPs.

## Description

Empty commerce tab (no fields only a text "Magento Store Configuration") is displayed in the page properties of storefront pages: "us", "en", "search".

The solution moves the 'Commerce' tab from the archetype to the aem-core-cif-components project and adds render conditions to the tab to be displayed only when it's needed.

Related PR: https://github.com/adobe/aem-cif-project-archetype/pull/103

## Related Issue

CIF-1321

## How Has This Been Tested?

Manually.

## Types of changes

## Checklist:


ui.apps/src/main/content/jcr_root/apps/core/cif/components/structure/page/v1/page/_cq_dialog/.content.xml
```diff
@@ -0,0 +1,80 @@
...
+                                            <categoryFilter jcr:primaryType="nt:unstructured"
+                                                sling:resourceType="commerce/gui/components/common/cifcategoryfield"
+                                                fieldDescription="Category ids for which this page will be used."
+                                                fieldLabel="Category ids for which this page will be used."
+                                                multiple="{Boolean}true"
+                                                name="./selectorFilter"
+                                                selectionId="id">
+                                                <granite:rendercondition jcr:primaryType="nt:unstructured"
+                                                    sling:resourceType="core/cif/components/renderconditions/pagetemplate"
+                                                    templatePath="/conf/atest/settings/wcm/templates/category-page"/>
```
dplaton, line 55: Who is `/conf/atest`?

ui.apps/src/main/content/jcr_root/apps/core/cif/components/structure/page/v1/page/_cq_dialog/.content.xml
```diff
@@ -0,0 +1,80 @@
...
+                                                fieldDescription="Product slugs for which this page will be used."
+                                                fieldLabel="Product slugs for which this page will be used."
+                                                multiple="true"
+                                                name="./selectorFilter"
+                                                rootPath="/var/commerce/products"
+                                                selectionId="slug">
+                                                <granite:rendercondition
+                                                    jcr:primaryType="nt:unstructured"
+                                                    sling:resourceType="core/cif/components/renderconditions/pagetemplate"
+                                                    templatePath="/conf/atest/settings/wcm/templates/product-page"/>
```
dplaton, line 68: Same question as above.

ui.apps/src/main/content/jcr_root/apps/core/cif/components/structure/page/v1/page/_cq_dialog/.content.xml
mhaack: Since we need the `templatePath` here, which we don't have, should we move the portions or the entire dialog to the archetype?

LSantha: @dplaton , good catch!
@mhaack , I'm looking for an alternative since this dialog would normally belong to this project.

LSantha: I've eliminated the page template from the render conditions.

mhaack: LGTM now

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/servlets/PageTemplateRenderConditionServlet.java
```diff
@@ -34,6 +34,8 @@
 /**
  * This servlet handles <code>granite:rendercondition</code> requests from the CIF page component,
  * in order to decide if the product and category pickers should be displayed in the page properties dialog.
+ *
+ * @deprecated
```
mhaack, line 38: Why no remove this Servlet? This is "internal" used, nothing customers directly dependent on. And if you replace it we don't need it. Customers which don't upgrade can still use the older version of the bundle.

LSantha: In principle this server is public API because it's exposed publicly via the resource type core/cif/components/renderconditions/pagetemplate. If a customer is using it somewhere in a similar way as it was used in the archetype then that code will break. 
Strictly speaking the new servlet is not a drop-in replacement for this because it's doing different kind of checks and it's not using templates, it's part of a different approach to solve the issue in the page properties dialog.

Nevertheless, if we decide to remove it, I'm also fine with that because the likelihood of somebody using it elsewhere may be quite small. I just thought that deprecation is the safer way.

mhaack: > Nevertheless, if we decide to remove it, I'm also fine with that because the likelihood of somebody using it elsewhere may be quite small. I just thought that deprecation is the safer way.
+1 for remove, the number of customers using this might be overseeable. Most might be even on CIF versions before it was introduced.

PR #258, opened 2020-04-28
CIF-1356 - Add caching support to the GraphQL client

cjelger:
This is the "companion" PR of https://github.com/adobe/commerce-cif-graphql-client/pull/14 to allow caching for all components and also for the faceted search OSGi service. Caching can then be enabled by simply configuring the GraphQL client with:

```
venia/components/structure/navigation:true:10:600
com.adobe.cq.commerce.core.search.services.SearchFilterService:true:10:600
```

here for example, to enable caching for the navigation component, and for the introspection queries done by the faceted search service.

Documentation will be added to the repo wiki when the PR is merged. Note that the build for this PR fails until https://github.com/adobe/commerce-cif-graphql-client/pull/14 is merged and released.

## How Has This Been Tested?

## Types of changes

## Checklist:


mhaack:
One thought on the `public static MagentoGraphqlClient create(Resource resource) {` vs. `public static MagentoGraphqlClient create(Resource resource, Page page) {` methods.

This is confusing for users because first (old) create method the resource was used to get the client, new method resource is used for something else and page is used to get the client.

I suggest we keep the old one (without deprecation tag) and also used that in the components. Only use the new one in template components. This will also reduce the # of changes in the PR, right.

cjelger:
> I suggest we keep the old one (without deprecation tag) and also used that in the components. Only use the new one in template components. This will also reduce the # of changes in the PR, right.

Yes I think it's better, I wasn't sure if we wanted to deprecate and change everything or keep the old API, so let's keep the PR small: I'll revert these changes.

cjelger:
@mhaack Done with reverting the changes. I only kept the changes for the navigation component, and also removed the caching done by the faceted search service so it's now done in the GraphQL client.

PR #274, opened 2020-05-18
CIF-1328: Refactor the Create Account process

laurentiumagureanu:
## Description
1. The shopper fills in the required data and clicks the "Create account" button.
2. The account is created and then the shopper is presented with a message `Your account was successfully created. You will receive a link at {{email}}. Access that link to confirm your email address.` and a `Sign in` button.
3. The shopper logs in.

## Related Issue

CIF-1328

## Motivation and Context

In our current implementation, the Create Account process automatically logs in the user after the account is created. This goes against the good practices in the industry because usually `Create Account` and `Sign in` are two separate user actions.

## How Has This Been Tested?

- unit tests
- manual functional testing

## Types of changes

## Checklist:


react-components/src/context/__test__/NavigationContext.test.js
```diff
@@ -0,0 +1,205 @@
...
+    it('shows sign in page', async () => {
+        const { getByText, getByTestId } = render(
+            <MockedProvider addTypename={false}>
+                <NavigationContextProvider>
+                    <ContextWrapper />
+                </NavigationContextProvider>
+            </MockedProvider>
+        );
+
+        expect(getByText('Sign in')).not.toBeUndefined();
```
herzog31, line 77: `expect(getByText('Sign in'))` should be sufficient, since the `getBy*` query throws an error if an element cannot be found. See https://testing-library.com/docs/dom-testing-library/api-queries#getby.

react-components/src/context/__test__/NavigationContext.test.js
```diff
@@ -0,0 +1,205 @@
...
+
+        fireEvent.click(getByText('Sign in'));
+        const result = await waitForElement(() => getByTestId('sign_in'));
```
herzog31, line 80: Same here. You could write:
```javascript
await waitForElement(() => getByTestId('sign_in')); // fails if element cannot be found after timeout
expect(result.textContent).toEqual('Sign in page');
```

react-components/src/context/__test__/NavigationContext.test.js
laurentiumagureanu: True but `expect` is rarely used without a matcher.
It also fails linting.

PR #298, opened 2020-06-11
Combine category query with product search query

laurentiumagureanu:
## Description
Combine category query with product search query in Category product list

## Related Issue

CIF-1172

## Motivation and Context

Performance optimization for faceted search

## How Has This Been Tested?

Unit tests
Manual tests

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/search/internal/services/SearchResultsServiceImpl.java
```diff
@@ -44,18 +47,7 @@
...
-import com.adobe.cq.commerce.magento.graphql.FilterRangeTypeInput;
-import com.adobe.cq.commerce.magento.graphql.Operations;
-import com.adobe.cq.commerce.magento.graphql.ProductInterface;
-import com.adobe.cq.commerce.magento.graphql.ProductInterfaceQuery;
-import com.adobe.cq.commerce.magento.graphql.ProductInterfaceQueryDefinition;
-import com.adobe.cq.commerce.magento.graphql.ProductPriceQueryDefinition;
-import com.adobe.cq.commerce.magento.graphql.ProductsQueryDefinition;
-import com.adobe.cq.commerce.magento.graphql.Query;
-import com.adobe.cq.commerce.magento.graphql.QueryQuery;
+import com.adobe.cq.commerce.magento.graphql.*;
```
cjelger, line 50: We don't use `*` imports as defined by the AEM developer guidelines/rules, please revert this.

bundles/core/src/main/java/com/adobe/cq/commerce/core/search/internal/services/SearchResultsServiceImpl.java
```diff
@@ -109,23 +113,23 @@ public SearchResultsSet performSearch(
...
+        String queryString = generateQueryString(searchOptions, availableFilters, productQueryHook, categoryRetriever);
         LOGGER.debug("Generated query string {}", queryString);
         GraphqlResponse<Query, Error> response = magentoGraphqlClient.execute(queryString);
 
         // If we have any errors returned we'll log them and return an empty search result
         if (response.getErrors() != null && response.getErrors().size() > 0) {
             response.getErrors().stream()
                 .forEach(err -> LOGGER.error("An error has occurred: {} ({})", err.getMessage(), err.getCategory()));
-            return searchResultsSet;
+            return new ImmutablePair<>(null, searchResultsSet);
```
cjelger, line 132: I think here you could still try to return the category (of course if there is any `data`), the errors might be related to something else. GraphQL can return data and errors together.

bundles/core/src/test/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImplTest.java
```diff
@@ -127,6 +127,7 @@ public void setUp() throws Exception {
         Utils.setupHttpResponse("graphql/magento-graphql-attributes-result.json", httpClient, HttpStatus.SC_OK, "{customAttributeMetadata");
         Utils.setupHttpResponse("graphql/magento-graphql-category-result.json", httpClient, HttpStatus.SC_OK, "{category");
         Utils.setupHttpResponse("graphql/magento-graphql-search-result.json", httpClient, HttpStatus.SC_OK, "{products");
+        Utils.setupHttpResponse("graphql/magento-graphql-search-result-with-category.json", httpClient, HttpStatus.SC_OK, "{products");
```
cjelger, line 130: These two rules are conflicting, they share the same prefix `{products`. I'm not sure what happens then, I mean which response you get. Actually I think you probably no longer need the previous lines with `magento-graphql-search-result.json` and `magento-graphql-category-result.json` because this is no longer needed (the components now does a single query for both = the new file you added). If these two files are no longer needed, please also remove them.

I think you should also extend the test to verify that the GraphQL client does the number of queries you expect (that is, one).

bundles/core/src/test/java/com/adobe/cq/commerce/core/components/testing/Utils.java
```diff
@@ -82,6 +82,28 @@ public static String setupHttpResponse(String filename, HttpClient httpClient, i
...
+     * @param httpClient The HTTP client for which we want to mock responses.
+     * @param httpCode The http code that the mocked response will return.
+     * @param startsWith When set, the body of the GraphQL POST request must start with that String.
+     * @param contains When set, the body of the GraphQL POST request must contain that String.
+     *
+     * @return The JSON content of that file.
+     *
+     * @throws IOException
+     */
+    public static String setupHttpResponse(String filename, HttpClient httpClient, int httpCode, String startsWith, String contains)
```
cjelger, line 105: I'm not entirely sure, but you don't need that at all, no? If not, please revert this change so we don't introduce something that we never use.

bundles/core/src/test/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImplTest.java
laurentiumagureanu: My bad, forgot to use `setupHttpResponse(String filename, HttpClient httpClient, int httpCode, String startsWith, String contains)` defined in `Utils.java`.
Will fix this

bundles/core/src/test/java/com/adobe/cq/commerce/core/components/testing/Utils.java
laurentiumagureanu: See the above comment

bundles/core/src/test/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImplTest.java
cjelger: Not sure you need this, because IIUC, you don't need the two previous response files any more.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImpl.java
```diff
@@ -159,12 +162,27 @@ public boolean showImage() {
...
+    private Pair<CategoryInterface, SearchResultsSet> getCategorySearchResultsSet() {
+        if (categorySearchResultsSet == null) {
+            Consumer<ProductInterfaceQuery> productQueryHook = categoryRetriever != null ? categoryRetriever.getProductQueryHook() : null;
+            categorySearchResultsSet = searchResultsService
+                .performSearch(searchOptions, resource, productPage, request, productQueryHook, categoryRetriever);
+        }
+        return categorySearchResultsSet;
+    }
+
+    public CategoryInterface getCategory() {
```
cjelger, line 179: Sorry I missed that, this does't have to be `public`, right?

bundles/core/src/main/java/com/adobe/cq/commerce/core/search/internal/services/SearchResultsServiceImpl.java
```diff
@@ -44,6 +47,7 @@
 import com.adobe.cq.commerce.core.search.services.SearchFilterService;
 import com.adobe.cq.commerce.core.search.services.SearchResultsService;
 import com.adobe.cq.commerce.graphql.client.GraphqlResponse;
+import com.adobe.cq.commerce.magento.graphql.*;
```
cjelger, line 50: Sorry I insist, we don't do `*` imports. ;-)

bundles/core/src/test/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImplTest.java
```diff
@@ -125,8 +125,7 @@ public void setUp() throws Exception {
 
         Utils.setupHttpResponse("graphql/magento-graphql-introspection-result.json", httpClient, HttpStatus.SC_OK, "{__type");
         Utils.setupHttpResponse("graphql/magento-graphql-attributes-result.json", httpClient, HttpStatus.SC_OK, "{customAttributeMetadata");
-        Utils.setupHttpResponse("graphql/magento-graphql-category-result.json", httpClient, HttpStatus.SC_OK, "{category");
```
cjelger, line 128: I think `magento-graphql-category-result.json` is no longer needed, you can use your new file in in the `setUp()` method to do `category = Utils.getQueryFromResource("graphql/magento-graphql-search-result-with-category.json").getCategory();`

bundles/core/src/main/java/com/adobe/cq/commerce/core/search/internal/services/SearchResultsServiceImpl.java
laurentiumagureanu: Argh, I fixed that but my IDE replaced it. I fixed it this time

PR #305, opened 2020-06-15
CIF-1369: Support for Preview with Product in page editor

hbacila:
A Product / Category page preview with real data can be triggered from outside the core components (addon, cif connector). We need a servlet which generates the preview url based on `UrlProvider` template using the provided parameters and redirects to that preview page.   

## Description

The page preview functionality is supported by a new servlet (`PreviewServlet`) bound to path `/bin/wcm/cif.html` with 2 possible selectors:`previewproduct` and `previewcategory`. The servlet will receive as parameters any of the following:

- `UrlProvider.ID_PARAM`
- `UrlProvider.SKU_PARAM`
- `UrlProvider.VARIANT_SKU_PARAM`
- `UrlProvider.URL_KEY_PARAM`
- `UrlProvider.VARIANT_URL_KEY_PARAM`

The page url used for preview is extracted from the `Referer` header. That represents the context of the preview (product page editor, category page editor).

## Motivation and Context

Previewing a page / category page with real data requires url formatting based on the UrlProvider template. We do not want to add dependency on cif-core-components bundle in addon and connector modules hence the support added to core components bundle. 

## How Has This Been Tested?

- unit testing
- functional testing 

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist:


hbacila:
@mhaack Currently the servlet checks if the referer contains `editor.html/`.  I can add extra checks for WCMMODE

mhaack:
Can we also check for the runmode, that should be the most save way I think?

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/servlets/PreviewServlet.java
```diff
@@ -0,0 +1,102 @@
...
+    property = {
+        "sling.servlet.methods=GET",
+        "sling.servlet.paths=/bin/wcm/cif",
+        "sling.servlet.selectors=" + PreviewServlet.PREVIEW_PRODUCT + ", " + PreviewServlet.PREVIEW_CATEGORY,
+        "sling.servlet.extensions=html"
+    })
+public class PreviewServlet extends SlingSafeMethodsServlet {
+    private static final Logger LOGGER = LoggerFactory.getLogger(PreviewServlet.class);
+
+    private static final String REFERER_HEADER = "Referer";
```
cjelger, line 49: Sorry I missed this one: I'd rather use `org.apache.http.HttpHeaders.REFERER` instead of introducing our own constant.

hbacila: fixed. thanks

PR #309, opened 2020-06-17
CIF-1464 - The FeaturedCategoryList component does not use the UrlPro…

cjelger:
…vider service

- replaced hard-coded String format with urlProvider

## How Has This Been Tested?

Extended unit test + manually tested.

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/categorylist/FeaturedCategoryListImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/categorylist/FeaturedCategoryListImpl.java
@@ -127,7 +127,10 @@ public class FeaturedCategoryListImpl implements FeaturedCategoryList {
         }
         List<CategoryTree> categories = categoriesRetriever.fetchCategories();
         for (CategoryTree category : categories) {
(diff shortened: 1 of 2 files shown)


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/categorylist/FeaturedCategoryListImpl.java
```diff
@@ -120,7 +127,8 @@ private void initModel() {
         }
         List<CategoryTree> categories = categoriesRetriever.fetchCategories();
         for (CategoryTree category : categories) {
-            category.setPath(String.format("%s.%s.html", categoryPage.getPath(), category.getId()));
+            Map<String, String> params = new ParamsBuilder().id(category.getId().toString()).map();
```
herzog31, line 130: `CategoriesRetriever` is also fetching the `url_path` (see https://github.com/adobe/aem-core-cif-components/blob/master/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/categorylist/CategoriesRetriever.java#L29-L33). So why not pass it to the url provider as well?

cjelger: Good point, I added that

PR #314, opened 2020-06-22
CIF-1436 - Can't remove or edit items from shopping cart

dplaton:
## Description

The problem was caused by the minicart refactoring that we've done some time ago - when the first product is added to the cart the "handlers" were not initialized. I removed the "cart handler" methods and use the actions directly.

The same problem affects the coupons feature - both "Add coupon" and "Remove coupon" were broken because of the same issue. The same fix has been applied to those as well.

## Related Issue

CIF-1436

## Motivation and Context

Fix the shopping cart functionality

## How Has This Been Tested?

Functional testing

## Types of changes

## Checklist:

### DIFF
--- a/react-components/src/components/Minicart/__test__/couponForm.test.js
+++ b/react-components/src/components/Minicart/__test__/couponForm.test.js
@@ -13,8 +13,11 @@
  ******************************************************************************/
 
 import React from 'react';
+import { MockedProvider } from '@apollo/react-testing';
 import { render, fireEvent } from '@testing-library/react';
 import { I18nextProvider } from 'react-i18next';
+
+import MUTATION_ADD_COUPON from '../../../queries/mutation_add_coupon.graphql';
 
 import { CartProvider } from '../cartContext';
 import CouponForm from '../couponForm';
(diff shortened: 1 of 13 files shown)


react-components/src/components/Minicart/__test__/couponForm.test.js
```diff
@@ -39,37 +44,13 @@ describe('<CouponForm />', () => {
...
+                    <CartProvider initialState={initialState} reducerFactory={() => state => state}>
+                        <CouponForm />
+                    </CartProvider>
+                </MockedProvider>
             </I18nextProvider>
         );
         expect(asFragment()).toMatchSnapshot();
     });
-
-    it('applies an coupon', () => {
```
herzog31, line 50: Since we don't have E2E tests, I would like to keep the tests that make sure that clicks on "Apply coupon" calls the `useCouponForm` resp. "Remove coupon" calls the `useCouponItem` hooks.

dplaton: I was thinking of implementing some tests for the `useXXX` "hooks" I added, but you're right, we still need to test the wiring here.

PR #320, opened 2020-06-25
CIF-1472 - Support Bundle Products: price display

cjelger:
- added the "From To" price display similar to how it's done in Magento Luma for bundle products
- refactored the JS/Karma unit tests to really use the PriceFormatter class and not a mock
- added JS support for i18n in the PriceFormatter

## How Has This Been Tested?

Extended unit tests and manually tested (see screenshots).

Screenshot 2020-06-25 at 15 30 14
Screenshot 2020-06-25 at 15 30 57

## Types of changes

## Checklist:


ui.apps/karma.conf.js
```diff
@@ -30,7 +30,6 @@ module.exports = function(config) {
     files: [
       'node_modules/@babel/polyfill/dist/polyfill.js',
       'src/main/content/jcr_root/apps/core/cif/components/commerce/product/**/js/*.js',
-      'src/main/content/jcr_root/apps/core/cif/components/commerce/productlist/**/js/*.js',
```
herzog31, line 33: I think was messed up a bit in #263. Not sure why this works or not works? 

For the productcollection component we have tests at https://github.com/adobe/aem-core-cif-components/blob/master/ui.apps/test/components/commerce/productcollection/productCollectionTest.js. Those tests require the clientlib at https://github.com/adobe/aem-core-cif-components/blob/master/ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/productcollection/v1/productcollection/clientlibs/js/productcollection.js.

cjelger: The file `productcollection.js` is directly imported in `productCollectionTest.js` so that's why it works. If we want to clean this up, we should create another issue as this is not related to my change (I removed that line because there was a warning in the build that there isn't any matching file with that path).

PR #323, opened 2020-06-26
CIF-1240 - Braintree validation error clear

laurentiumagureanu:
## Description

clears previous validation error when changing providers

## Related Issue

CIF-1240

## Motivation and Context

Braintree error message stays after changing to other Braintree payment method.

## How Has This Been Tested?

Manually

## Types of changes

## Checklist:


react-components/src/components/Checkout/paymentsForm.js
```diff
@@ -173,31 +173,42 @@ const PaymentsForm = props => {
...
-                </div>
-                {billingAddressFields}
-            </div>
-            <div className={classes.footer}>
-                <Button onClick={cancel}>{t('common:cancel', 'Cancel')}</Button>
-                <Button priority="high" type="submit" disabled={isSubmitting}>
-                    {t('checkout:use-payment-method', 'Use Payment Method')}
-                </Button>
-            </div>
+            {({ formApi }) => (
```
herzog31, line 176: Ideally we refactor this to use the hook instead. It's fine for now. Let's do that as part of the customisation effort.

dplaton: > Let's do that as part of the customisation effort.

@herzog31, what customization effort are we talking about?

I'd say we do this now since using the `useFormApi()` hook makes the cod more readable

herzog31: You would need to refactor the whole component. That's a bit much for the scope of a simple bug fix in my opinion.

laurentiumagureanu: @dplaton Accessing the formApi via hooks doesn't work as expected.
Modified the code to use it like so:

```javascript
import { Form, useFormApi } from 'informed';

const PaymentsForm = props => {
...
const formApi = useFormApi();
return (
        <Form className={classes.root} initialValues={initialFormValues} onSubmit={handleSubmit}>
            <div className={classes.body}>
                <h2 className={classes.heading}>Billing Information</h2>
                <div className={classes.braintree}>
                    <Select
                        items={paymentMethodsItems}
                        field="payment_method"
                        initialValue={paymentMethod}
                        onValueChange={() =>
                            formApi.getError('payment_nonce') && formApi.resetField('payment_nonce')
                        }
                    />
                </div>
....
```

and it results in an error at runtime:
```
Uncaught TypeError: formApi.getError is not a function
    at onValueChange (react-components.js:87966)
    at Object.setValue (react-components.js:40671)
    at handleChange (react-components.js:41324)
    at HTMLUnknownElement.callCallback (react-components.js:128949)
    at Object.invokeGuardedCallbackDev (react-components.js:128998)
    at invokeGuardedCallback (react-components.js:129053)
    at invokeGuardedCallbackAndCatchFirstError (react-components.js:129067)
    at executeDispatch (react-components.js:129197)
    at executeDispatchesInOrder (react-components.js:129222)
    at executeDispatchesAndRelease (react-components.js:129326)
```

herzog31: `useFormApi` can only be used in a component that's a child of the component that renders `<Form>`. So you cannot use it in `PaymentsForm`. That's why I said using this hook would require a complete refactoring of the component.

Issue #303, opened 2020-06-15
My account dropdown - move the Sign in option in the dropdown


dplaton:
### User Story
As a visitor, I want to have a dedicated space from which I can access "Sign in" options, without having to open the navigation side panel. 

### Deliverables
1. Updated Header component with the icon for the user dropdown
2. The drop-down containing only the sign-in button, with complete functionality (enter username and password, sign in)
4. The sign-out option must be accessible and functional
3. After the sign in the dropdown will only display the user name and the sign-out option.

### Acceptance Criteria

1. The feature is implemented
2. The unit tests meet the coverage criteria
3. The UI fits the Venia theme

### Verification Steps
1. Create a new project using the archetype
2. Install the SNAPSHOT version of this project.
3. Check the home page:
1.1 The user icon must be present in the header component
1.2 Clicking the icon opens the dropdown which has the sign-in form
1.3 Enter an existing username and password and click "Sign in" - the user should be signed in and the "Sign out" button should be displayed
1.4 Click the "Sign out" option. It should work as expected.

### Mockups:
Before logging in: 

- we should keep the "Sign in" text next to the icon
<img width="527" alt="Screenshot 2020-06-23 at 10 34 33" src="https://user-images.githubusercontent.com/1407821/85374392-61ffad80-b53d-11ea-9876-b0f875e90662.png">

- after login, dropdown closed

<img width="428" alt="Screenshot 2020-06-23 at 10 34 54" src="https://user-images.githubusercontent.com/1407821/85374404-64fa9e00-b53d-11ea-9f75-f92012a29e3b.png">

- dropdown open
- we don't need all the links here, but we should at least have "Account information" and "Address Book"
<img width="467" alt="Screenshot 2020-06-23 at 10 35 13" src="https://user-images.githubusercontent.com/1407821/85374407-65933480-b53d-11ea-8dcc-bd2f8c5ca111.png">

### Internal ref
CIF-1444 and CIF-1445

mhaack:
Issue resolved and verified.

PR #326, opened 2020-07-02
CIF-1444 - [React Component] Move the sign-in component in the user drop-down

dplaton:
## Description

This PR adds an account icon and its dropdown on the header of the venia storefront, this account dropdown will provide the same features that implemented in navigation side panel at the moment, including sign-in, reset the password, create a new account, changing the password and sign out after sign-in. 

## Related Issue

The PR is related to issue #303 

## Motivation and Context

We now have account-related features available in the navigation side panel, but in order to access those features, a user has to open the side panel. We want users to have direct access via an account dropdown that is toggled by clicking an account icon added to the storefront header. In the meantime, we want to hide account-related features from the navigation side panel when it's desktop view, and display account icon added to header instead, on mobile view, we will hide the account icon on the header and show those features in the side panel.

## How Has This Been Tested?

This PR has been tested by adding the tests to render the expected snapshots to make sure the markup is not changed accidentally. 

image

image

image

image

image

## Types of changes

## Checklist:


react-components/src/components/AccountContainer/__test__/accountContainer.test.js
```diff
@@ -0,0 +1,48 @@
...
+import { render } from '@testing-library/react';
+
+import UserContextProvider from '../../../context/UserContext';
+import { CartProvider } from '../../Minicart/cartContext';
+import i18n from '../../../../__mocks__/i18nForTests';
+
+import AccountContainer from '../accountContainer';
+
+describe('<AccountContainer>', () => {
+    beforeAll(() => {
```
herzog31, line 27: Please review the way you use mocking here. Make sure that the mocked method is restored after the test. See for example https://jestjs.io/docs/en/mock-functions#mocking-modules.

react-components/src/components/AccountContainer/accountContainer.js
```diff
@@ -0,0 +1,47 @@
...
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { useTranslation, Trans } from 'react-i18next';
+
+import { useUserContext } from '../../context/UserContext';
+
+import AccountTrigger from './accountTrigger';
+import AccountDropdown from './accountDropdown';
+
+const container = document.querySelector('.header__accountTrigger #miniaccount');
```
herzog31, line 23: For re-useability, can the selector be a prop on the component and the query is done inside the component?

react-components/src/components/AuthBar/container.js
```diff
@@ -21,7 +21,7 @@ import { useNavigationContext } from '../../context/NavigationContext';
 import classes from './container.css';
 
 const Container = () => {
-    const container = document.querySelector('#miniaccount');
+    const container = document.querySelector('aside.navigation__root #miniaccount');
```
herzog31, line 24: Same like above :)

react-components/src/components/AccountContainer/accountContainer.js
dplaton: I think we'd better pass all the selectors in a config object for the whole app

herzog31: Should live in https://github.com/adobe/aem-cif-guides-venia/pull/12 then, so it can be easier customized by customers :)

react-components/src/components/AccountContainer/__test__/accountContainer.test.js
nettree: Hi @herzog31, are we saying the mockClear() here like:
```
    afterEach(() => {
        ReactDOM.createPortal.mockClear();
    });
```
?

react-components/src/components/AccountContainer/accountContainer.js
nettree: So do we want to deal with this later by making changes in https://github.com/adobe/aem-cif-guides-venia? @herzog31 @dplaton ?

react-components/src/components/AccountContainer/__test__/accountContainer.test.js
herzog31: Please check https://jestjs.io/docs/en/mock-function-api#mockfnmockclear. `mockClear` doesn't restore the original functionality. It resets the recorded calls. I'm referring to restore the original functionality :) 

Try
```
jest.mock('react-dom');
ReactDOM.createPortal.mockImplementation(e => e);
```

nettree: Hi @herzog31, I see what you mean, I tried with jest.mock('react-dom'), but then I receive the error reported by test script, it says as below:
image
So I can't mock the whole `react-dom` module. I tried to use `spyOn`, which works quite well, so we can have:
```
    beforeAll(() => {
        // mock createPortal because we don't have the DOM element to render the AccountDropdown
        jest.spyOn(ReactDOM, 'createPortal').mockImplementation(element => element);
    });
```
Are you happy with this?
The other consideration is, I actually get the current implementation from `react-components/src/components/CartTrigger/__test__/cartTrigger.test.js`, not sure we should change the implementation there as well to keep consistent?

herzog31: That works as well 👍 
We can fix the implementation in the other test later.

react-components/src/components/AccountContainer/accountContainer.js
nettree: Hi @herzog31, the query selector is now passed via `<CommerceApp>`, please review :)

react-components/src/components/AuthBar/container.js
nettree: Hi @herzog31, same here, the query selector is now passed via `<CommerceApp>`, please review

herzog31: Looks much better :)

react-components/src/components/AccountContainer/__test__/accountContainer.test.js
nettree: I've changed the `accountContainer.test.js` to use `jest.spyOn()` then :)

nettree:
Hi @herzog31, @dplaton, there is no further change I'm going to commit to this PR, so please let me know if you want me to change anything else. Thank you.

react-components/src/context/UserContext.js
```diff
@@ -114,11 +128,14 @@ const UserContextProvider = props => {
...
         isSignedIn: isSignedIn(),
+        isAccountDropdownOpen: false,
         signInError: null,
         inProgress: false,
         createAccountError: null,
         createAccountEmail: null,
-        cartId: null
+        cartId: null,
+        accountDropdownView: 'SIGN_IN',
+        accountContainerQuerySelector
```
dplaton, line 138: @nettree Somehow I missed this, sorry for the late feedback: I don't think the `accountContainerQuerySelector` belongs in the user state. This is more like an "application state" element.

We should create an `ApplicationContext` that wraps all the app and makes this configuration available. I'll create a github issue.

nettree: Yea, I agree, I was struggling a bit with this as well but can't find a more suitable place at that time, `ApplicationContext` sounds good, do you want me to add it? Or we want to deal with it later?

dplaton: We'll dewl with that in a separate issue

PR #330, opened 2020-07-07
CIF-1484 - Panel title not updated when returning to sign-in

LSantha:
Make sure that navigation panel title is always updated when pressing the back arrow on the panel.

## Description

The navigation panel title is not updated after visiting the forgot password or create account panels and pressing the back arrow in the top of the navigation panel.

## Related Issue

CIF-1484

## How Has This Been Tested?

Manually.

## Types of changes

## Checklist:

- [] I have added tests to cover my changes and the overall coverage did not decrease.

### DIFF
--- a/react-components/src/actions/navigation/actions.js
+++ b/react-components/src/actions/navigation/actions.js
@@ -92,9 +92,9 @@ export const showAccountCreated = ({ dispatch, t }) => {
 };
 
 export const showView = ({ dispatch, t, view }) => {
-    const title = stepTitles[view](t);
+    const title = stepTitles[view];
     if (title) {
-        dispatchEvent(new CustomEvent('aem.accmg.step', { detail: { title: title } }));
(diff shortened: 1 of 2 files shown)


react-components/src/actions/navigation/actions.js
```diff
@@ -90,3 +90,11 @@ export const showAccountCreated = ({ dispatch, t }) => {
     dispatchEvent(new CustomEvent('aem.accmg.step', { detail: { title: stepTitles[view](t) } }));
     dispatch({ type: 'changeView', view });
 };
+
+export const showView = ({ dispatch, t, view }) => {
+    const title = stepTitles[view](t);
+    if (title) {
```
herzog31, line 96: If `view` doesn't exist in the `stepTitles` object, there will already be an error thrown. So the `if (title)` is too late. Should be like:
```
let title = stepTitles[view];
if (!title) return;
title = title(t);
// ...
```

LSantha: Thanks. I've fixed it.

PR #331, opened 2020-07-08
CIF-1485 - The cq:catalogPath restriction not applied to picker when previewing non-empty page

LSantha:
Fixed PreviewServlet to work with non empty PDP and PLP.

## Description

Preview servlet cannot handle PDP and PLP which is already showing real data.
I've noticed this issue while working on https://github.com/adobe/commerce-cif-connector/pull/137 and I fixed it because it would be an obstacle for its verification.

## Related Issue

CIF-1485

## How Has This Been Tested?

Manually, JUnit.

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/servlets/PreviewServlet.java
```diff
@@ -93,10 +92,7 @@ private Page getRefererPage(SlingHttpServletRequest request) {
 
         int idx = referer.lastIndexOf(PAGE_EDITOR_PATH);
         String pagePath = referer.substring(idx + PAGE_EDITOR_PATH.length());
-        pagePath = pagePath.replaceAll(".html", ""); // get rid of .html extension
-
-        ResourceResolver resourceResolver = request.getResourceResolver();
-        Resource pageResource = resourceResolver.getResource(pagePath);
-        return pageResource != null ? pageResource.adaptTo(Page.class) : null;
+        Resource pageResource = request.getResourceResolver().resolve(pagePath);
+        return pageResource.adaptTo(Page.class);
```
mhaack, line 96: Why is the null check removed here? There might be some issue with the request parameter and pagePath can be empty or wrong.

LSantha: ResourceResolver.resolve() differs from ResourceResolver.getResource() by returning NonExistentResource instead of null and then adaptTo(Page.class) will return null.

PR #335, opened 2020-07-13
CIF-1416 - Add CIF Core Components to AEM Components Library Page

cjelger:
- fix issues for deployment on AEM as a Cloud Service

## Description

This fixes some remaining issues when deploying the components to AEMaaCS:
- a transitive dependency for `graphql-java` is no longer exported in AEMaaCS
- the GraphQL endpoint must be configured with an environment variable

## How Has This Been Tested?

Tested on AEMaaCS with SNAPSHOT dependencies for the examples library.

## Types of changes

## Checklist:

### DIFF
--- a/examples/ui.apps/src/main/content/jcr_root/apps/cif-components-examples/config/com.adobe.cq.commerce.graphql.client.impl.GraphqlClientImpl-examples.config
+++ b/examples/ui.apps/src/main/content/jcr_root/apps/cif-components-examples/config/com.adobe.cq.commerce.graphql.client.impl.GraphqlClientImpl-examples.config
@@ -1,5 +1,5 @@
 identifier="examples"
-url="https://localhost:8443/apps/cif-components-examples/graphql"
+url="$[env:CIF_MOCK_GRAPHQL_ENDPOINT]"
 httpMethod="GET"
 acceptSelfSignedCertificates="true"
 maxHttpConnections="20"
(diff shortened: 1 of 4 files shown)


examples/ui.apps/src/main/content/jcr_root/apps/cif-components-examples/config/com.adobe.cq.commerce.graphql.client.impl.GraphqlClientImpl-examples.config
```diff
@@ -1,5 +1,5 @@
 identifier="examples"
-url="https://localhost:8443/apps/cif-components-examples/graphql"
+url="$[env:CIF_MOCK_GRAPHQL_ENDPOINT]"
```
mhaack, line 2: Why not using `COMMERCE_ENDPOINT` here? Yes it is a mock GraphQl server not a real on, but with using COMMERCE_ENDPOINT we can also deploy this together with the CIF Add-On.

cjelger: I'd really keep this separate, mainly because the mock server currently only supports GraphQL requests for the components (I have no idea what responses you'd get for example for the Omnisearch), and also to really keep that separate from the production configuration. Once can easily set the two variables, but I'd not mix a mock config with a production config.

cjelger:
@mhaack To verify, install the examples library on AEM Cloud SDK and check that the bundle starts (it doesn't start without that fix). You can also test the `CIF_MOCK_GRAPHQL_ENDPOINT` endpoint variable.

PR #347, opened 2020-07-24
CIF-1498 - Setup IT project for CIF components

cjelger:
This PR adds some basic integration tests for the CIF components based on the CIF components library.

The tests are rather basic and only check that each component renders some "basic" elements like for example the product name, the products of a category, etc, without going into all the possible details. The scope of each test could be extended if needed, the main goal here is to ensure that each component properly renders and all elements (components, graphql client, etc, work together).

Note that the `it/http` project is ignored when we perform a release by setting the `skip-it` Maven profile during the release.

## Checklist:

### DIFF
--- a/.circleci/ci/build.js
+++ b/.circleci/ci/build.js
@@ -1,0 +1,32 @@
+/*******************************************************************************
+ *
+ *    Copyright 2019 Adobe. All rights reserved.
+ *    This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License. You may obtain a copy
+ *    of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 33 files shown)


herzog31:
Just noticed, but might be unrelated to this PR. The integration tests correctly report the `integration` flag for codecov. The flags for `unittests`, `karma` and `jest` are somehow not correctly applied. Maybe we should update the orb or use the bash uploader directly.

cjelger:
@herzog31 I fixed the codecov flags by setting the orb to the `1.1.1` latest version.

herzog31:
@cjelger Yes that helped 👍 . All flags are available now. See for example https://codecov.io/gh/adobe/aem-core-cif-components/src/6f2142699eb5aa56e24dfe6317da4a681b58f9ba/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java.

PR #352, opened 2020-07-28
CIF-1529 - Include CIF Add-On in IT setup for Cloud SDK

cjelger:
**Important**: the URL to download the CIF Add-On is set as a CircleCI variable `CIF_ADDON_URL`. This is obfuscated in the build output so nobody can have access to the add-on.

This has the main drawback that the version of the Add-On is also not visible, so we might forget to update it with each new release of the Add-On. Also because variables are also obfuscated in the CircleCI backend, it might become impossible with time to know which version we currently use.

We could hence maybe rename the variable to `CIF_ADDON_URL_2020_7_7_SNAPSHOT` so it would be straightforward from the name what version we use. If we release a new version of the Add-On, we would add for example the variable `CIF_ADDON_URL_2020_7_28` and update the build.

The drawback here is that we'd have to also update the variable in the `it-tests.js` file.

Feedback or other ideas is welcome.

### DIFF
--- a/.circleci/ci/it-tests.js
+++ b/.circleci/ci/it-tests.js
@@ -29,17 +29,21 @@ try {
         // Connect to QP
         ci.sh('./qp.sh -v bind --server-hostname localhost --server-port 55555');
 
-        // The core components are already installed in the Cloud SDK
-        let extras = '';
+        // We install the graphql-client by default except with the CIF Add-On
+        let extras = '--bundle com.adobe.commerce.cif:graphql-client:1.6.1:jar';
         if (process.env.AEM == 'classic') {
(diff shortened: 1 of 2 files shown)


cjelger:
@mhaack I think we could keep `integration-test-cloudready` so that we test the components on Cloud SDK without the Add-On. In case `integration-test-cloudready-with-addon` fails, we'd know that the problem comes from the Add-On and not the Cloud SDK. In case both fail, we'd know the problem comes from the Cloud SDK. WDYT?

mhaack:
Ok I'm fine with that it might help to find and delimit errors.

Issue #327, opened 2020-07-03
[My Account] Address book management


dplaton:
### User Story

As a registered shopper, I want to manage my shipping/billing addresses so I can use them properly during the checkout process.

### Description & Motivation

Improve the checkout experience by allowing the shopper to select a previously stored address in the checkout process.

### Deliverables

1. A dedicated AEM page for address book management (can be created using the "Content Page" template), accessible via the "Address book" link from the account drop down or side-panel.

2. A self contained React component that allows address book management:

<img width="1217" alt="address_book_3" src="https://user-images.githubusercontent.com/1407821/86461711-10bc9e80-bd33-11ea-8786-1e8e570a94f7.png">

The component must allow the developer to customize the display type (list/grid) via props or a similar mechanism.

Each entry in the list must have the "Edit" and "Remove" actions:

<img width="479" alt="address_book_delete_confirmation" src="https://user-images.githubusercontent.com/1407821/86462709-b7556f00-bd34-11ea-9db1-c358484db500.png">

Adding a new address must be done via the same form used in the checkout process

<img width="760" alt="address_book_2" src="https://user-images.githubusercontent.com/1407821/86462764-d3f1a700-bd34-11ea-9403-81482b818563.png">

3. Unit tests for the React component

### Acceptance Criteria

1. The Address book page can be accessed via the link in the account dropdown or side panel
2. The address book page is working as expected - you can create, update, or delete addresses.
3. You can select one and only one address to be the "default" address.

PR #353, opened 2020-07-29
CIF-1398 - [My Account] Address book management

nettree:
This PR is to add the feature of address book management for the signed in user

## Description

* User can see the saved address by accessing the address book page

* User can create a new address, edit and delete an existing address

* User can set an address to their default address(shipping and billing at the moment)

## Related Issue
#327 

## Motivation and Context
This is to enhance the feature set of the user account so that the user can manage their shipping and billing addresses when they've signed in.

## How Has This Been Tested?
All the changes made have jest tests covered, features are tested in the browser, including create, edit, delete the address, and set the default address

## Types of changes

## Checklist:


react-components/src/context/NavigationContext.js
```diff
@@ -72,6 +75,10 @@ const NavigationContextProvider = props => {
 
     const showAccountCreated = () => NavigationActions.showAccountCreated({ dispatch, t });
 
+    const showAddressBook = () => {
+        window.location.href = navigationState.addressBookPath;
```
herzog31, line 79: Shouldn't we rather use `history.push` here? See https://reactrouter.com/web/api/Hooks/usehistory.

nettree: currently, the `react-router-dom` is only installed in venia repo not in core components repo, are you happy with how the router is used in venia repo? I tested in my local, it seems that browser controls history works as well, but I will have a look for sure, thank you @herzog31 :)

dplaton: We can install `react-router-dom` in `aem-core-cif-components`, but we have to put it as `peerDependency`. But as long as the feature works without it...

herzog31: Browser history should work with `window.location.href`, just wondering if it has any side effects in terms of navigating within React. Probably only works for navigating between actual AEM pages.

dplaton:
@nettree I quickly check the feature and run into some bugs:

1. You can create several "default" addresses. There should be only one default address
2. When you edit an address the e-mail field is empty.

nettree:
Hi @dplaton, this is something I'd like to discuss with you, it seems like the endpoints don't accept email as one of the address field, in this case, shall we hide the email field from the address form for address book?

react-components/src/context/NavigationContext.js
nettree: Hi @herzog31, I tried to use `useHistory` and then `history.push`, but since we don't have React route context introduced in core components repo yet, it doesn't seem working at the moment, so are you happy for leaving this logic as is at the moment or do you have other thoughts on this? :)

herzog31: If React navigation works with it, let's keep it :)

nettree: Thanks for that @herzog31 :)

PR #366, opened 2020-08-06
CIF-1567 - Client-side GraphQL should be validated against Magento schema

cjelger:
This PR ensures that all client-side GraphQL requests are validated against the Magento schema. We only test the "raw" requests declared *without parameters* in `.graphql` files. I'd have liked to also test the queries sent with parameters by each component test, but didn't find any way to do that with Apollo's `MockedProvider`.

While working on this, I detected an invalid request in `CommerceGraphqlApi` which was actually no longer used so I removed it.

**Note**: I intentionally copied the `magento-schema-2.3.5.json` file twice in each sub-project, to avoid having one project depend on the other for testing.

## How Has This Been Tested?

Locally tested: changing a query with an invalid field makes the corresponding test fail.

## Checklist:


herzog31:
Officially in version 1.2.0 we support Magento versions 2.3.4 and 2.3.5. Would it make sense to test against both schemas in the unit tests?

cjelger:
My intention was not really to test all supported versions, in the bundle we also only really get the "validation" with the latest library version. The goal is more to make sure we detect breaking changes that could be introduced in the Magento schema (like the removal of `createBraintreeClientToken` in 2.4.0).

herzog31:
But shouldn't that go both ways? We want to detect breaking changes introduced by Magento. Also we want to make sure that a change in our code is compatible with the schema. And this is especially true if we want to support multiple versions. Imagine we add a new feature (something my account) that was added with 2.3.5, but is not available in 2.3.4. If nobody actually tests it on 2.3.4, we might forget to declare that 2.3.4 is no longer supported because of that feature.

herzog31:
`jest` supports [describe.each](https://jestjs.io/docs/en/api#describeeachtablename-fn-timeout) and [it.each](https://jestjs.io/docs/en/api#testeachtablename-fn-timeout) for parameterised tests. I think this would make everything a bit more read-able :)

cjelger:
@herzog31 I now used `describe.each` and `it.each`, thx for the feedback.

PR #367, opened 2020-08-07
CIF-1536 - Product teaser: Add CTA label configuration

cjelger:
- add label field in dialog

The change in `.content.xml` looks large because of reformatting, the main change is at the bottom with the extra `ctaText` field.

Screenshot 2020-08-07 at 11 39 21

## Checklist:

### DIFF
--- a/ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/productteaser/v1/productteaser/_cq_dialog/.content.xml
+++ b/ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/productteaser/v1/productteaser/_cq_dialog/.content.xml
@@ -63,7 +63,7 @@
                                                 jcr:primaryType="nt:unstructured"
                                                 sling:resourceType="granite/ui/components/coral/foundation/form/textfield"
                                                 fieldDescription="The text displayed by the call-to-action button"
-                                                fieldLabel="Call to action: text"
+                                                fieldLabel="Call to action text"
                                                 name="./ctaText"/>
                                         </items>
                                     </column>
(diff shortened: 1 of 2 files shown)


ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/productteaser/v1/productteaser/_cq_dialog/.content.xml
```diff
@@ -1,61 +1,70 @@
...
+                                                        jcr:primaryType="nt:unstructured"
+                                                        text="See more details"
+                                                        value="details"/>
                                                 </items>
                                             </call-to-action>
+                                            <ctaText
+                                                jcr:primaryType="nt:unstructured"
+                                                sling:resourceType="granite/ui/components/coral/foundation/form/textfield"
+                                                fieldDescription="The text displayed by the call-to-action button"
+                                                fieldLabel="Call to action: text"
```
mhaack, line 66: `action:` why is there a colon here in the label?

ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/productteaser/v1/productteaser/actions.html
```diff
@@ -17,11 +17,11 @@
     <div class="productteaser__cta" data-sly-test="${product.callToAction != null}">
         <button data-sly-test="${product.callToAction == 'add-to-cart'}" data-action="addToCart" data-item-sku="${product.sku}"
                 class="button__root_highPriority button__root clickable__root button__filled" type="button">
-            <span class="button__content"><span>Add to Cart</span></span>
+            <span class="button__content"><span>${properties.ctaText @ i18n}</span></span>
```
mhaack, line 20: I18n for something entered by the user? Will that even work?

Also there is no default fallback. If the user adds the component to the page but not enters a call to action text this will render an empty button. I suggest to use something like `${properties.text || 'Add to Cart' @ i18n}`.

herzog31:
Don't we break Sling model delegation if we use the property directly in HTL? Let's say a customer would want to add a prefix to the call to action text?

mhaack:
> Don't we break Sling model delegation if we use the property directly in HTL? Let's say a customer would want to add a prefix to the call to action text?
She/he can just type in the prefix or if it should be permanent the project can overlay the HTL.

ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/productteaser/v1/productteaser/actions.html
cjelger: Sure `i18n` will work, as long as the text entered by the author is in the translation dictionaries. I'm fine to add the default.

ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/productteaser/v1/productteaser/_cq_dialog/.content.xml
cjelger: Just because I added it for the label: I can remove it, no problem.

cjelger:
@mhaack I included the changes you proposed.

PR #373, opened 2020-08-11
CIF-1416 - Add CIF Core Components to AEM Components Library Page

cjelger:
To be able to deploy the CIF components library on top of the deployment project for the WCM components library, we need a single `all` content package for the examples, where we embed all the CIF dependencies.

To be able to deploy `-SNAPSHOT` versions on its `dev` environment, the WCM deployment project checks out all repos to install `-SNAPSHOT` artifacts. So to only deploy the `examples-all` project, we set the `<cloudManagerTarget>none</cloudManagerTarget>` property to all content packages except the `examples-all`.

To add this property, we use the `filevault-package-maven-plugin` instead of the `content-package-maven-plugin`.

## How Has This Been Tested?

Tested on the `dev` environment of the WCM components library project on AEMaaCS.

## Checklist:


herzog31:
It seems weird to me to set configurations in the non example packages for deployment of the component library. But I guess the only option we have to prevent this is to deploy snapshot versions of the component artifacts somewhere, right?

PR #376, opened 2020-08-13
CIF-1546 - Pass the DOM nodes in which the components render as props

herzog31:
## Description

* Created new `ConfigContext` which provides mounting point selectors, page paths and GraphQL configurations.
* Removed all portals from the project and moved them to the Venia project.
* Created `Portal` component which can be used to wrap components to render them in a given selector.

See https://github.com/adobe/aem-cif-guides-venia/pull/33.

## How Has This Been Tested?

* Updated unit tests
* Manual test with Venia project

## Types of changes

## Checklist:

### DIFF
--- a/react-components/src/context/UserContext.js
+++ b/react-components/src/context/UserContext.js
@@ -23,6 +23,7 @@ import {
     signOutUser as signOutUserAction,
     deleteAddress as deleteAddressAction
 } from '../actions/user';
+import { useConfigContext } from './ConfigContext';
 
 import MUTATION_DELETE_CUSTOMER_ADDRESS from '../queries/mutation_delete_customer_address.graphql';
 import MUTATION_REVOKE_TOKEN from '../queries/mutation_revoke_customer_token.graphql';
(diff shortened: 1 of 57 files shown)


react-components/src/context/UserContext.js
```diff
@@ -217,12 +218,7 @@ const UserContextProvider = props => {
     const [userCookie, setUserCookie] = useCookieValue('cif.userToken');
     const [, setCartCookie] = useCookieValue('cif.cart');
     const isSignedIn = () => !!userCookie;
-    const accountContainerQuerySelector =
-        (props.config && props.config.accountContainerQuerySelector) || '.header__accountTrigger #miniaccount';
-    const addressBookContainerQuerySelector =
-        (props.config && props.config.addressBookContainerQuerySelector) || '#addressbook';
-    const addressBookPath =
-        (props.config && props.config.addressBookPath) || '/content/venia/us/en/my-account/address-book.html';
+    const { pagePaths } = useConfigContext();
```
dplaton, line 221: A context depending on other context is a bad idea IMO. I understand why it needs to be done here, but we need to refactor the UserContext anyway.

herzog31: Actually it's a common pattern. But I agree. Just look at all the tests I had to adapt :D 

We would basically need to move https://github.com/adobe/aem-core-cif-components/blob/c7ff96a7a879d518d4c79b3b7795966b786b8710/react-components/src/context/UserContext.js#L316-L318 somewhere else. Maybe we can remove it from the context and have every component that uses it implement it by taking the path from the new `ConfigContext`.

dplaton: I think that's a better approach. UserContext holds too much information as it is.

herzog31: Okay, I refactored both `UserContext` and `NavigationContext` to not depend on `ConfigContext` anymore :)

PR #378, opened 2020-08-17
CIF-1473: Support Bundle Products: option selection

laurentiumagureanu:
## Description

**ProductOptionsBundle** react component will be mounted in a placeholder div element that is present for bundle products.
Initially displays a **Customize** button. Once the button is pressed, it performs a graphQL query to fetch the bundle options and renders it accordingly.

The add to cart is not implemented as there's a separate issue for it.

## Related Issue

CIF-1473

## Motivation and Context

Support Bundle Products

## How Has This Been Tested?

Unit tests; manually

image
image

## Types of changes

## Checklist:


react-components/src/components/BundleProductOptions/bundleProductOptions.js
```diff
@@ -0,0 +1,265 @@
...
+                                                <polyline points="6 9 12 15 18 9"></polyline>
+                                            </svg>
+                                        </span>
+                                    </span>
+                                </span>
+                                <p className="message-root"></p>
+                            </div>
+                        </section>
+                        <section className="productFullDetail__cartActions productFullDetail__section">
+                            <button
```
herzog31, line 242: The event handler on the add to cart button via https://github.com/adobe/aem-core-cif-components/blob/master/ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/product/v1/product/clientlib/js/addToCart.js is only done once for when the whole document is loaded. Since this component starts with an initial `bundleState = null`, the button is not rendered and the event handler should not be set.

dplaton:
I'm fine with the whole code (I have more or less the same comments as @herzog31), but I have a problem with the whole approach: all our "React Core Components" are standalone. They do share context, but you can use the shopping cart without the authentication part, and you can use the whole My Account suite of components without the Shopping Cart. 

However, this feature is tied to a Product component. Even more, this component doesn't even share state with the other components, it's just used for a specific flavor of a product component. With this in mind, I'm really wondering if we should even have this component as part of the React Core Components project. I can think of two options to think about:

1/ Have it as a React component, but part of `ui.apps` (like we used to have the "Venia Demo")
2/ Just have it in plain JS that uses the GraphQL API, just like the price component.

laurentiumagureanu:
> I'm fine with the whole code (I have more or less the same comments as @herzog31), but I have a problem with the whole approach: all our "React Core Components" are standalone. They do share context, but you can use the shopping cart without the authentication part, and you can use the whole My Account suite of components without the Shopping Cart.
> 
> However, this feature is tied to a Product component. Even more, this component doesn't even share state with the other components, it's just used for a specific flavor of a product component. With this in mind, I'm really wondering if we should even have this component as part of the React Core Components project. I can think of two other options:
> 
I agree that having this one functionality as part of `React Core Components` is a bit overkill and the use frequency will be low. 

> 1/ Have it as a React component, but part of `ui.apps` (like we used to have the "Venia Demo")
>
I think I can just create a **React** app as part of the **Product** `clientlib` and have it mounted just for Bundle Products

> 2/ Just have it in plain JS that uses the GraphQL API, just like the price component.
>
Having this implemented as plain Javascript can be hard to do and maintain, the main issue being the HTML rendering (which can easily become spaghetti code).

herzog31:
IMO having yet another React project in our repos is overkill. I'm completely fine with having this feature as part of the `react-components` project. Maybe you could refactor the way the component gets the `sku`, so it could be (unlikely) used in other contexts as well.

dplaton:
> I think I can just create a React app as part of the Product clientlib and have it mounted just for Bundle Products

That's a good idea since the client libraries already get built with Webpack, so no need to create another React project. The only downside is that you'd have to re-write your tests.

> IMO having yet another React project in our repos is overkill. 
That's my thinking as well, @herzog31, hence the idea above.

react-components/src/components/BundleProductOptions/bundleProductOptions.js
laurentiumagureanu: the intention is to not use the default event handler for adding to cart since a bundle requires a different mutation.
I can modify the classes to avoid any confusion.

laurentiumagureanu: but that would mean copying the CSS rules from `productFullDetail__cartActions` to another class and use that one instead in this button. not sure it's worth it

dplaton (approved):

react-components/src/components/BundleProductOptions/__test__/bundleProductOptions.test.js
```diff
@@ -0,0 +1,84 @@
...
+                    <BundleProductOptions />
+                </MockedProvider>
+            </I18nextProvider>
+        );
+
+        expect(asFragment()).toMatchSnapshot();
+    });
+
+    it('renders the component with sku', () => {
+        jest.spyOn(document, 'querySelector').mockImplementation(selector => {
```
herzog31, line 47: You could just add an outer `<div id="bundle-product-options" data-sku="sku">` to the render method.

react-components/src/components/BundleProductOptions/__test__/bundleProductOptions.test.js
```diff
@@ -0,0 +1,84 @@
...
+import { render } from '@testing-library/react';
+import * as ConfigContext from '../../../context/ConfigContext';
+
+import i18n from '../../../../__mocks__/i18nForTests';
+
+import BundleProductOptions from '../bundleProductOptions';
+
+describe('<BundleProductOptions>', () => {
+    beforeAll(() => {
+        // mock createPortal because we don't have the DOM element to render the BundleProductOptions
```
herzog31, line 26: Please update the outdated comment.

react-components/src/components/BundleProductOptions/bundleProductOptions.js
```diff
@@ -0,0 +1,258 @@
...
+                option.customization.reduce((a, c) => {
+                    return a + c.price * c.quantity;
+                }, 0)
+            );
+        }, 0);
+
+        return <Price currencyCode={currencyCode} value={price} className="bundlePrice" />;
+    };
+
+    if (sku) {
```
herzog31, line 172: You could simplify all of these conditions into:
```
if (!sku) return <></>
if (bundleState === null) return button
return all options
```

react-components/src/components/BundleProductOptions/checkbox.js
```diff
@@ -0,0 +1,68 @@
...
+
+const Checkbox = props => {
+    const { item, customization, sortedOptions, handleSelectionChange } = props;
+
+    const onChange = event => {
+        const { checked, value } = event.target;
+        let newCustomization;
+
+        if (checked) {
+            newCustomization = customization;
```
herzog31, line 26: Are you sure this works as expected? You copy a reference to an array here, so any change made to `newCustomization` is also applied to `customization`. Probably better to us
```
newCustomization = [...customization];
```

herzog31 (commented):
I would like so see some tests for all the `onX` methods of the input fields (checkbox, multiselect, radio etc).

react-components/src/components/BundleProductOptions/checkbox.js
laurentiumagureanu: doesn't really matter. changing `customization` does not trigger rendering. and `newCustomization` is then applied to a function coming from the parent component which triggers a state change thus triggering re-rendering.
So, this works as expected but it's better to do a shallow copy to avoid any issues/confusion

herzog31 (approved):

PR #379, opened 2020-08-19
CIF-1551 - [React Components] Refactor the NavigationContext

dplaton:
## Description

A React Context is a useful feature, allowing components to access context data without having the receive it via props. However, this comes with a cost - the components need to be wrapped in the context provider. We must always weigh the pros and cons of using a context and judge on a case by case basis.

In this particular case, the state of the "My Account Panel" (renamed it!) can be managed by the local state without the need of a "Context". I implemented a `useNavigationState` hook that manages all the states of the My Account panel and I removed the `NavigationContext` and the actions module. Also, I updated the `AuthBar` component so that it uses the new state management.

## Related Issue

CIF-1551

## Motivation and Context

Reduce code complexity.

## How Has This Been Tested?

New unit tests, functional testing.

## Types of changes

## Checklist:


react-components/src/components/AuthBar/useNavigationState.js
```diff
@@ -0,0 +1,124 @@
...
+    CHANGE_PASSWORD: t => t('account:change-password', 'Change Password'),
+    MY_ACCOUNT: t => t('account:my-account', 'My account'),
+    ACCOUNT_CREATED: t => t('account:account-created', 'Account created'),
+    SIGN_IN: t => t('account:sign-in', 'Sign In')
+};
+
+const startAccMgEvent = new CustomEvent(events.START_ACC_MANAGEMENT);
+const exitAccMgEvent = new CustomEvent(events.EXIT_ACC_MANAGEMENT);
+
+const navigationPanel = document.querySelector('aside.navigation__root');
```
herzog31, line 46: The selector should be taken from the `ConfigContext`, so it can be customized.

react-components/src/queries/__test__/validateQueries.test.js
```diff
@@ -31,6 +31,7 @@ describe.each([
     });
 
     expect(files.length).toBeGreaterThan(0); // Ensures we read the right folder
+    // eslint-disable-next-line no-console
     console.log(`Validating ${files.length} GraphQL requests against Magento schema ${version}`);
```
herzog31, line 35: Please remove this line instead of setting an `eslint` exception.

cjelger: It was really useful to have that line in the tests, to see at least that something happens and that multiple versions are being processed.

dplaton: @cjelger the Magento version information is already displayed as part of the test description

herzog31 (approved):

PR #380, opened 2020-08-20
CIF-1548 - Generate CSS API file

herzog31:
## Description

* Added script to generate CSS "API" file which includes all classes but no styles. This can be used as starting point for styles in a new project.
* Generated files are called `*-template.css`.
* The file is published as part of the npm package.
* Documentation updated in https://github.com/adobe/aem-cif-guides-venia/pull/41.

## Types of changes

## Checklist:


react-components/css-template.js
```diff
@@ -0,0 +1,38 @@
...
+    .forEach(f => {
+        let cssPath = path.join(distFolder, f);
+
+        // Clear css classes
+        let css = fs.readFileSync(cssPath, { encoding: 'utf-8' });
+        let emptyCss = css.replace(/\{([^{}]*)\}/gm, '{}');
+
+        // Store in new file
+        let { dir, name, ext } = path.parse(cssPath);
+        let cssPathNew = path.format({ dir, ext, name: `${name}-template` });
```
dplaton, line 36: Maybe we should put the templates in a subfolder, like `css-api` or something similar.

dplaton (approved):
We need to document this feature in the README - what are CSS API files and how to use them.

react-components/package.json
```diff
@@ -16,7 +16,7 @@
   "scripts": {
     "lint": "eslint 'src/**/*.js'",
     "webpack:dev": "webpack --mode=development",
-    "webpack:prod": "webpack --mode=production",
+    "webpack:prod": "webpack --mode=production && node css-template.js",
```
cjelger, line 19: I'm not sure what you exactly want to achieve: only generate the files when we release (but they wouldn't be committed in the repo, right?) so that they are included in the NPM package? (sorry I haven't checked the recent changes to create the NPM package).

herzog31: The templates are created in the `dist` folder which is published as npm package as part of the release. It's also generated for normal builds, but since the `dist` folder is part of `.gitignore`, it will never be committed.

cjelger: Aren't "normal builds" building with `webpack:dev`? Also if we do that for React Components, do we also want to the same for the `css` styles of the HTL components?

herzog31: > Aren't "normal builds" building with webpack:dev?
Correct :)

> Also if we do that for React Components, do we also want to do the same for the css styles of the HTL components?
Those files already exist in https://github.com/adobe/aem-cif-guides-venia/tree/main/ui.frontend/src/main/styles. We could automate the generation of those files too, but it would work differently, since we would need to commit them directly to the venia project + archetype.

cjelger (approved):

PR #393, opened 2020-09-04
[React Components] Increase coverage for the Editable Form component

alkim91:
Added snapshot / unit tests for Editable Form Component

## Description

Added unit tests to editableForm component in order to improve coverage.

## Related Issue

https://jira.corp.adobe.com/browse/CIF-1601

## Motivation and Context

According to our code coverage reports, the coverage for EditableForm fell quite low to ~38%.  This is the first attempt to improve the coverage here.

## How Has This Been Tested?

This PR doesn't introduce any changes to non-testing code, I tested it using `npm run test` from the react-components repository, verifying that all unit tests & snapshot tests pass.

Screen Shot 2020-09-04 at 2 50 02 PM

## Types of changes

## Checklist:


alkim91 (commented):
@dplaton  @herzog31 @mhaack  

My first PR attempt. Looks like this actually dropped the coverage for some reason :( 

I want to test the business logic in editableForm OTHER than the switch case statements(The event handlers and hooks basically) - but I couldn't figure out a way to reliably spy on hooks. Is there a good example of this in our repo?

dplaton:
@alkim91 what hooks do you want to spy on? Generally, you don't need to spy on hooks, rather you want to assert on their effects - for example, check that an `useEffect` hook has done its job when the component has mounted by inspecting the component after mount.

react-components/src/components/Checkout/__test__/editableForm.test.js
```diff
@@ -114,63 +99,157 @@ describe('<EditableForm />', () => {
...
+                            __typename: 'AvailableShippingMethod'
+                        }
+                    ],
+                    selected_shipping_method: null,
+                    __typename: 'ShippingCartAddress'
+                }
+            ],
+            available_payment_methods: [
+                {
+                    code: 'braintree_paypal',
```
herzog31, line 138: I think you should ignore all Braintree logic here for now, since you're testing the displaying of the payment methods form.

react-components/src/components/Checkout/__test__/__snapshots__/editableForm.test.js.snap
```diff
@@ -1,3 +1,343 @@
...
+        >
+          Use Address
+        </span>
+      </button>
+    </div>
+  </form>
+</DocumentFragment>
+`;
+
+exports[`<EditableForm /> should render ShippingForm if user is entering shipping information 1`] = `<DocumentFragment />`;
```
herzog31, line 341: Why is this empty? Shouldn't it display the `ShippingForm` component?

react-components/src/components/Checkout/__test__/__snapshots__/editableForm.test.js.snap
```diff
@@ -1,3 +1,343 @@
...
+
+exports[`<EditableForm /> should render the payments form if user is editing payments 1`] = `<DocumentFragment />`;
```
herzog31, line 343: Why is this empty? Shouldn't it display the `PaymentsForm` component?

react-components/src/components/Checkout/__test__/__snapshots__/editableForm.test.js.snap
alkim91: You're right. Taking a second look.

react-components/src/components/Checkout/__test__/__snapshots__/editableForm.test.js.snap
alkim91: You're right. Taking a second look.

PR #396, opened 2020-09-09
CIF-1624: Remove commons-text dependency in UrlProviderImpl

hbacila:
## Description

This PR replaces the use of `StringSubstitutor.replace()` in `UrlProviderImpl` with `StringUtils` methods to avoid adding dependency on Apache's `commons-text` package.

## Motivation and Context

The `commons-text` bundle was embedded in `ui.apps` package. Embedding 3rd party bundles in application type package is now prohibitted for AEM as a Cloud Service.

## How Has This Been Tested?

- unit tests

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/services/UrlProviderImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/services/UrlProviderImpl.java
@@ -20,7 +20,6 @@ import java.util.Map;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
-import org.apache.commons.text.StringSubstitutor;
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.apache.sling.api.resource.Resource;
 import org.osgi.service.component.annotations.Activate;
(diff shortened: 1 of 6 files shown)


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/services/UrlProviderImpl.java
```diff
@@ -162,4 +161,24 @@ private String parseIdentifier(IdentifierLocation identifierLocation, SlingHttpS
             throw new RuntimeException("Identifier location " + identifierLocation + " is not supported");
         }
     }
+
+    static class StringSubstitutor {
```
mhaack, line 165: Instead of inventing our own new class can't we just use https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#replaceEach-java.lang.String-java.lang.String:A-java.lang.String:A- ? This bundle is already available in AEM.

Or as alternative use a regex?

hbacila: Our `StringSubstitutor:replace` method uses just that (`StringUtils.replaceEach`). Our own `StringSubstituor` was created to preserve the `UrlProviderImpl` implementation.

mhaack: I saw we use StringUtils.replaceEach, wonder if that also can be done simpler? Mabe with String.formart() or RegEx. But I'm just nitpicking here.

LSantha (approved):

herzog31 (approved):

PR #399, opened 2020-09-16
CIF-1469 - CIF Core Components support for staged catalog data

cjelger:
This PR adds the support for staged catalog data for Magento and AEM Launches. If the current page is inside an AEM Launch, a `Preview-Version` HTTP header will be sent with all GraphQL requests. The "epoch" value of the header is taken from the live date of the AEM Launch.

This also requires that a Magento integration token (with permission "Catalog > Staging") must be configured as an `Authorization:Bearer <token>` HTTP header in the GraphQL client.

Also important: I adapted the navigation and breadcrumb components so they use the "production version" of the current page to get their configuration, an also adapted the `SiteNavigation` class to make sure that it creates the "correct" links for the product and category pages. Here "correct" means that if the Launch includes the product and/or category page(s), then the links should point to the pages from the Launch and not to production pages.

## How Has This Been Tested?

Extended unit tests and extensively tested manually.

## Types of changes

## Checklist:


herzog31 (approved):
Looks good. What about the client-side code?

cjelger:
@herzog31 Good point, I'll create a follow-up issue for that and add it for technical grooming.

PR #405, opened 2020-09-23
CIF-1084 - Implement the password reset page

herzog31:
## Description

This PR depends on #400.

* Add new `ResetPassword` React component to display the reset password form.
* Add new `resetpassword` component to display the placeholder for the React component.
* Added unit tests.
* This feature requires setup on the Magento side, please see https://github.com/adobe/aem-core-cif-components/wiki/Configure-Reset-Password.

Please also see https://github.com/adobe/aem-cif-guides-venia/pull/60.

## How Has This Been Tested?

* Manually tested password reset workflow.
* Unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/react-components/src/components/ResetPassword/useResetPassword.js
+++ b/react-components/src/components/ResetPassword/useResetPassword.js
@@ -1,0 +1,41 @@
+/*******************************************************************************
+ *
+ *    Copyright 2019 Adobe. All rights reserved.
+ *    This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License. You may obtain a copy
+ *    of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software distributed under
+ *    the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ *    OF ANY KIND, either express or implied. See the License for the specific language
+ *    governing permissions and limitations under the License.
+ *
+ ******************************************************************************/
+
+import { useState, useCallback } from 'react';
+import { useMutation } from '@apollo/react-hooks';
+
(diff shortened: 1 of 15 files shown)


react-components/src/components/ResetPassword/useResetPassword.js
```diff
@@ -0,0 +1,41 @@
...
+
+import MUTATION_RESET_PASSWORD from '../../queries/mutation_reset_password.graphql';
+
+const useResetPassword = () => {
+    const [status, setStatus] = useState('new');
+
+    const [resetPassword] = useMutation(MUTATION_RESET_PASSWORD);
+
+    const handleFormSubmit = useCallback(
+        async ({ email, token, password }) => {
```
cjelger, line 26: I think you should revalidate the email address here, and not just rely on the client-side validation which can easily be skipped, it's green-belt security ;-)

I'm also wondering if we can also use a CSRF token to make sure that nobody can just DDoS the form or do some kind of brute-force attack against Magento's reset password feature.

herzog31: > I think you should revalidate the email address here, and not just rely on the client-side validation which can easily be skipped, it's green-belt security ;-)

Validating the email again in JS is again client-side validation and can easily be skipped. No point in doing that twice ;)

> I'm also wondering if we can also use a CSRF token to make sure that nobody can just DDoS the form or do some kind of brute-force attack against Magento's reset password feature.

Good point. Currently we don't do this at all (login for example). I think this should be investigated in a separate issue.

cjelger: > Validating the email again in JS is again client-side validation and can easily be skipped. No point in doing that twice ;)

Good point, my brain was still in client-server mode ;-)

For CSRF I agree we could do this in a separate issue.

cjelger (approved):

dplaton (approved):

PR #409, opened 2020-09-28
CIF-1400 - MyAccount - Account Information Page

dplaton:
## Description

This PR adds the implementation of the `AccountDetails` components along with a few helper components. As a first, I included the GraphQL queries in the component's folders instead of the `./queries` folder since that folder already grew too large.

I added a `Dialog` component which renders the edit form in a modal dialog. This component was directly taken from Venia.
I added a `Password` field component to render the password field which features a "toggle" that shows the password in clear case.

PR in Venia: https://github.com/adobe/aem-cif-guides-venia/pull/63

## Related Issue

CIF-1400

## Motivation and Context

As a shopper, I want to update my personal data such as name, e-mail, and password.

## How Has This Been Tested?

Prerequisites: fresh instance of AEM (on-prem or cloud)

1/ Install Venia with the development version of AEM Core CIF React Components
2/ Login with an existing user
3/ Click the "Account information" link.
4/ Check that the Account Details page opens correctly and retrieves the correct data
5/ Click the "Edit" button
6/ Check that the modal dialog opens and it's pre-filled with the correct data.
7/ Change any of the data (either name or e-mail) and click "Save". 
8/ Check that the data is refreshed.
9/ Click the Edit button again
10/ Change the password - check that you get the validation error when you try to use the same password
11/ Click "Save"
12/ Sign out and sign in again with the new password - it should work.

Screenshot 2020-09-28 at 11 44 49

Screenshot 2020-09-28 at 11 44 43
## Types of changes

## Checklist:


alkim91 (approved):

herzog31:
I think if you move the queries out of the `./queries` folder, it makes things like https://github.com/adobe/commerce-cif-graphql-integration-reference/pull/30 much harder. So for now I would prefer to keep it in one place. @cjelger wdyt?

cjelger:
@herzog31 Definitely, and we also validate all queries with https://github.com/adobe/aem-core-cif-components/blob/master/react-components/src/queries/__test__/validateQueries.test.js so let's keep them all in that same folder.

dplaton:
@herzog31 @cjelger It's done, I moved the queries to the `queries` folder

react-components/src/components/AccountDetails/useAccountDetails.js
```diff
@@ -0,0 +1,113 @@
...
+ *    the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ *    OF ANY KIND, either express or implied. See the License for the specific language
+ *    governing permissions and limitations under the License.
+ *
+ ******************************************************************************/
+import { useMemo, useCallback, useState } from 'react';
+import { useQuery, useMutation } from '@apollo/react-hooks';
+import { useUserContext } from '../../context/UserContext';
+
+const useAccountDetails = props => {
```
herzog31, line 18: Since this hook stores state, I think it would be appropriate to use a React context here. So you don't need to pass all the exported logic and state here via props to child components like `EditModal` and `EditForm`.

react-components/src/utils/hooks/useScrollIntoView.js
```diff
@@ -0,0 +1,31 @@
...
+ *    the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ *    OF ANY KIND, either express or implied. See the License for the specific language
+ *    governing permissions and limitations under the License.
+ *
+ ******************************************************************************/
+import { useEffect } from 'react';
+
+const OPTIONS_DEFAULTS = { behavior: 'smooth', block: 'center' };
+
+/**
```
herzog31, line 18: Shouldn't we add a note or copyright header here, since this is taken from PWA studio?

react-components/src/components/AccountDetails/useAccountDetails.js
dplaton: There are only a few components under this tree, I don't see any reason to use a React Context just to manage the state of a modal. In general, I don't see any problem with passing state down to child components for trivial matters. I see React Contexts for cross-cutting concerns, like make some "context data" available to components that are not part of the same sub-tree.

react-components/src/utils/hooks/useScrollIntoView.js
dplaton: The original component doesn't have a copyright header - https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/hooks/useScrollIntoView.js
I'll add a link to the original code

herzog31: I think that's a problem actually. IIUC Magento uses OSL 3.0 which requires to publish any used code under OSL 3.0 as well which is incompatible with our Apache 2.0 license. I think we need to double check that.

dplaton: Yes, they use OSL 3.0. 
I'll revert the change and use a different method.

herzog31 (approved):

PR #412, opened 2020-09-30
CIF-1653 - TTL-based caching is skipped by the SpecificPageFilterFactory

cjelger:
- configure filter ranking to lowest possible value, to avoid skipping the DispatcherCacheHeaderFilter filters from ACS-Commons

## Motivation and Context

When it matches, the `SpecificPageFilterFactory` forwards the request to another page, and this "breaks" the chain of filters. We hence set the ranking to the lowest possible value, in particular to avoid skipping the TTL-based filters.

## How Has This Been Tested?

Manually tested with dispatcher and ACS-Commons filters.

Screenshot 2020-09-30 at 12 59 25

## Types of changes

## Checklist:


mhaack (approved):
Not sure if `Integer.MIN_VALUE` is the best here but I trust you tested this out ;-)

cjelger:
@mhaack I added a screenshot, the `Integer.MIN_VALUE` is used by other filters that don't need to be processed with any high priority, and yes it works fine for us.

mhaack:
>and yes it works fine for us.

I was expecting this :-) I was just curious.

PR #415, opened 2020-10-01
Extend Core Components to send data to client-data layer

laurentiumagureanu:
## Description

Send CIF Components data and events to client-data layer when the integration is enabled.

## Related Issue

CIF-1181

## Motivation and Context

Expose the default OOTB elements related to CIF Core Components

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:


hbacila:
LGTM.

One comment though: It would be probably better to move the data layer specific classes (`com.adobe.cq.commerce.core.components.models.datalayer,` `com.adobe.cq.commerce.core.components.models.internal.v1.datalayer`) outside the models package.
There was a similar request for WCM core components [here](https://github.com/adobe/aem-core-wcm-components/pull/1115#issuecomment-700630833) although it's too late in that case.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/categorylist/FeaturedCategoryListImpl.java
```diff
@@ -73,6 +74,7 @@
     private Map<String, Asset> assetOverride;
     private Page categoryPage;
     private AbstractCategoriesRetriever categoriesRetriever;
+    private List<CategoryTree> categories;
```
herzog31, line 77: You don't need this. The `retriever` classes will automatically populate and store the result once you call the `fetchCategories` method.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/ProductImpl.java
```diff
@@ -387,4 +387,40 @@ private String safeDescription(ProductInterface product) {
...
+    // DataLayer methods
+
+    @Override
+    public ComponentData getComponentData() {
+        return new ProductDataImpl(this, resource);
+    }
+
+    @Override
+    protected String generateId() {
+        return StringUtils.join("product", ID_SEPARATOR, productRetriever.fetchProduct().getId());
```
herzog31, line 399: Why do you use the `slug` for the productlistitem, but the `id` for the product component?

herzog31 (approved):
Looks good 👍  

Don't forget to also update https://github.com/adobe/aem-cif-guides-venia/blob/main/core/src/main/java/com/venia/core/models/commerce/MyProductTeaserImpl.java.

hbacila (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/ProductImpl.java
cjelger: I think we should completely avoid using `id` because in Magento this is basically the row id of the database entry, so nothing reusable or meaningful. Better use the `sku` everywhere, because the `slug` can change over time while the `sku` should never change.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/datalayer/CategoryData.java
```diff
@@ -0,0 +1,29 @@
...
+ *    OF ANY KIND, either express or implied. See the License for the specific language
+ *    governing permissions and limitations under the License.
+ *
+ ******************************************************************************/
+package com.adobe.cq.commerce.core.components.models.datalayer;
+
+public interface CategoryData {
+
+    default String getId() {
+        return null;
```
cjelger, line 19: Sometimes you have the `default` method returning null, sometimes throwing an UnsupportedOperationException, is this intentional, shouldn't this be consistent and everywhere the same?

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/common/ProductListItemImpl.java
```diff
@@ -107,4 +113,43 @@ public String getFormattedPrice() {
...
+
+    @Override
+    public String getId() {
+        String prefix = StringUtils.join(parentId, ID_SEPARATOR, ITEM_ID_PREFIX);
+        return StringUtils.join(prefix, ID_SEPARATOR, StringUtils.substring(DigestUtils.sha256Hex(getSlug()), 0, 10));
+    }
+
+    @Override
+    public String getDataLayerType() {
+        return "core/cif/components/commerce/productListItem";
```
cjelger, line 132: Shouldn't this be lowercase? We usually have lower case for resource types.

cjelger (changes requested):
Looks good, but I could not see a single unit tests for all the datalayer classes. You get some coverage (with integration tests?), but you never actually test that the generated JSON or returned data matches what you expect.

cjelger:
@laurentiumagureanu I added one unit test that to show what we should have for all components, as agreed on Slack, please add the missing ones for the other components and we are good to go! Thx.

cjelger (approved):

dplaton (approved):

PR #416, opened 2020-10-02
CIF-1629 - Enrich PLP / PDP automatically with metadata

cjelger:
- activates Sling model caching for the `Product` and `ProductList` components
- introduces the `PageMetadata` interface to provide meta data in a generic manner

Note that this PR is currently blocked by https://github.com/apache/sling-org-apache-sling-models-impl/pull/20 (a bug prevents model caching from working as expected).

   !!!   **DO NOT MERGE** before the new version of the Sling Model has been released and added to AEM   !!!

I haven't yet introduced any "canonical URL" tag, I'd like to discuss this again with the team.

## How Has This Been Tested?

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/page/PageMetadataImpl.java
```diff
@@ -0,0 +1,101 @@
...
+        if (metaKeywords == null && currentPage instanceof com.adobe.cq.wcm.core.components.models.Page) {
+            metaKeywords = ((com.adobe.cq.wcm.core.components.models.Page) currentPage).getKeywords().toString();
+        }
+        return metaKeywords;
+    }
+
+    @Override
+    public String getMetaTitle() {
+        String metaTitle = provider != null ? provider.getMetaTitle() : null;
+        return metaTitle != null ? metaTitle : currentPage.getTitle();
```
herzog31, line 77: Here you return the meta title if available, but otherwise the product name:
https://github.com/adobe/aem-core-cif-components/blob/973747b528bbae6fad5e6e2bda0ef1a2fa48014e/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/ProductImpl.java#L405-L408

So the page title is actually never used. Do we really want that? I would assume we go for the meta title if available and otherwise some kind of concatenation of page title and product name. Maybe we should check back with our PMs :)

herzog31 (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/page/PageMetadataImpl.java
mhaack: I think it is ok. For PDP/PLP we agreed that commerce is the source. The page title will still be used for all other pages using the page component.

mhaack (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/page/PageMetadataImpl.java
cjelger: Exactly, this is what we discussed/agreed. Customers would have to either override the model and these methods if they want to change this, and they can also override the `head.html` file and for example add a suffix to all pages with the sites name, to have something like `Chaz Hoodie - Venia Store`.

dplaton (approved):

PR #419, opened 2020-10-08
CIF-1657: [React Components] Clean up the unit tests warnings 

hbacila:
The unit tests for the React Components module all pass, but they throw a lot of warnings regarding missing props and the ever present `"Warning: An update to XXXX inside a test was not wrapped in act(...)."`

## Description

This change fixes the warnings above. It also fixes a styling issue for `AccountDetails/editForm` due to empty `editForm.css`

## Motivation and Context

We must cleanup these warnings as they can potentially hide some implementation issues.

## How Has This Been Tested?

- running unit test
- verify if components render correctly

## Types of changes

## Checklist:


react-components/src/actions/user/__test__/actions.test.js
```diff
@@ -17,6 +17,9 @@ const setCartCookie = jest.fn();
 const setUserCookie = jest.fn();
 const dispatch = jest.fn();
 
+// avoid console errors logged during testing
+console.error = jest.fn();
```
herzog31, line 21: I would rather add the `--silent` parameter to the `jest` call than mocking the `console.error` calls here. This has the advantage that you can easily decide whether or not you want to print the console statements. Sometimes they are useful.

herzog31 (approved):

react-components/src/actions/user/__test__/actions.test.js
hbacila: The problem with `--silent` is that it supresses all console logs including the warnings that may indicate an issue (such the ones we fixed here).

react-components/src/components/AuthBar/__test__/useNavigationState.test.js
```diff
@@ -18,13 +18,17 @@ import React from 'react';
...
+import { I18nextProvider } from 'react-i18next';
+import i18n from '../../../../__mocks__/i18nForTests';
 
 describe('useNavigationState', () => {
     const renderWithContext = (Component, props) =>
         render(
-            <ConfigContextProvider config={configObject}>
-                <Component {...props} />
-            </ConfigContextProvider>
+            <I18nextProvider i18n={i18n}>
```
dplaton, line 27: Since this the root of our context I propose we decorate `render` so that it includes it automatically (like in https://github.com/adobe/aem-core-cif-components/blob/master/react-components/src/utils/test-utils.js#L105).
Same with `ConfigContextProvider`.

dplaton (changes requested):
Simplify the mock components so that they won't require the `I18nextProvider` and the `ConfigContextProvider` every time. For the latter, we can always provide the full configuration.
We can do the same for the `MockedProvider`, but we need to allow passing the `mock` array to the render function

hbacila:
Refactored the unit tests so they all use the custom render (from `utils/test-utils.js`) which wraps the mock component with `MockedProvider`, `ConfigContextProvider`, `UserContextProvider` and `I18nextProvider`. The render's `options` parameter is extended with the following props:
- `config` - provides configuration object for `ConfigContextProvider`
- `userContext` - provides initialState for `UserContextProvider`
- `mocks` - provides mocks data for `MockedProvider`

react-components/src/context/__test__/UserContext.test.js
```diff
@@ -895,13 +804,7 @@ describe('UserContext test', () => {
...
         };
 
-        const { getByTestId } = render(
-            <MockedProvider mocks={mocks} addTypename={false}>
-                <UserContextProvider>
-                    <ContextWrapper />
-                </UserContextProvider>
-            </MockedProvider>
-        );
+        const { getByTestId } = render(<ContextWrapper />, { mocks: mocks });
```
dplaton, line 807: You can write `{mocks}` instead of `{mocks: mocks}` - shorthand notation in ES6

dplaton (approved):
Great work! This would make the tests so much more readable.

PR #420, opened 2020-10-12
CIF-1477 - Support Bundle Products: add example in CIF components library

cjelger:
This PR adds a `ui.frontend` React project to the CIF components examples folder, which builds the React app used to display the bundle product options. It also adds the support in the GraphQL Servlet used by the CIF components library.

The `ui.frontend` project was generated with the archetype 24, and then modified to remove as much as possible everything that was not required by the project.

## How Has This Been Tested?

Screenshot 2020-10-12 at 09 58 58

## Types of changes

## Checklist:


dplaton:
Nice work!

I think the whole typescript support can be dropped, but that depends on how close to the original archetype project we want to keep our example app.

dplaton (approved):

examples/ui.frontend/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "typescript": "^3.3.3333",
+    "watch": "^1.0.2",
+    "webpack": "^4.27.1",
+    "webpack-cli": "^3.1.2",
+    "webpack-dev-server": "^3.9.0",
+    "webpack-import-glob-loader": "^1.6.3",
+    "webpack-merge": "^4.2.1"
+  },
+  "dependencies": {
+    "@adobe/aem-core-cif-react-components": "^1.4.0",
```
herzog31, line 60: Might not be that important, since we only use this for one React component. The examples project uses the same version of dependencies as built (including snapshots), while the React components dependency points to the release. In the build of `react-components`, we already do an `npm link`, so maybe we could add an `npm link @adobe/aem-core-cif-react-components` to the build here as well to make sure, we use the exact same version as in the components build?

herzog31 (approved):

examples/ui.frontend/package.json
cjelger: Good point, I actually tried last week that but it didn't work ... the build then failed with some weird webpack error. I can create a follow-up error for that, that PR is long enough. ;-)

mhaack (approved):

PR #438, opened 2020-11-11
CIF-1637 - Support CIF Core Components staged catalog data with Timewarp

cjelger:
- adds timewarp support in MagentoGraphqlClient
- adapts all components to be able to use the timewarp feature
- extends unit tests and re-enables some old tests and coverage

## How Has This Been Tested?

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java
@@ -20,9 +20,12 @@ import java.util.Calendar;
 import java.util.List;
 import java.util.TimeZone;
 
+import javax.servlet.http.Cookie;
+
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.Header;
 import org.apache.http.message.BasicHeader;
+import org.apache.sling.api.SlingHttpServletRequest;
 import org.apache.sling.api.resource.Resource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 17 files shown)


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java
```diff
@@ -160,24 +187,53 @@ private MagentoGraphqlClient(Resource resource, Page page) {
...
+            requestOptions.withHttpMethod(HttpMethod.POST);
+        }
+
         if (!headers.isEmpty()) {
             requestOptions.withHeaders(headers);
         }
     }
 
+    private Long getTimeWarpEpoch(SlingHttpServletRequest request) {
+        String timeWarp = request.getParameter("timewarp");
```
dplaton, line 222: Is this the only place we use the request? If so, maybe we can refactor a bit to just get the `timewarp` as a parameter when you create the `MagentoGraphqlClient` instead of the full SlingHttpServletRequest object.

herzog31 (approved):

dplaton (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java
cjelger: Well if one day we need something else from the request, then we wouldn't have to change the `create` method API any more, so I'd prefer that we keep the request, also so that "users" of the method do not have to extract the parameter by themselves.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java
```diff
@@ -94,17 +100,38 @@ public static MagentoGraphqlClient create(Resource resource) {
      *            This is needed because it is not possible to get the current page for components added to the page template.
      *            If null, the resource will be used to adapt to the client, but this might fail for components defined on page templates.
      * @return A new MagentoGraphqlClient instance.
+     * @deprecated Use {@link #create(Resource, Page, SlingHttpServletRequest)} to be able to use the Timewarp feature to preview staged
+     *             data.
      */
+    @Deprecated
```
mhaack, line 106: Not sure if we should deprecate this constructor we can keep both.

mhaack (approved):

PR #439, opened 2020-11-11
CIF-1699 - [React Components] Clean-up the dependencies of the NPM module

dplaton:
## Description

* Convert the remaining class components to functional components
* Remove the unused plugins from the Babel configuration. Those plugins are no longer needed since we don't have class components anymore
* Remove the dependencies from `package.json`

## Related Issue

CIF-1699
## Motivation and Context

Clean up the dependency tree

## How Has This Been Tested?

Install the SNAPSHOT dependency using Venia and check that the client-side components still work (mainly the ones that use forms)

## Types of changes

## Checklist:


react-components/package.json
```diff
@@ -29,21 +29,15 @@
   "devDependencies": {
```
herzog31, line 29: Should we also sort the production and dev dependencies? IMO the following dependencies should be production or peer dependencies:
* `@apollo/client`
* `braintree-web-drop-in`
* `graphql`
* `i18next`
* `informed`
* `prop-types` ?
* `react`
* `react-dom`
* `react-feather`
* `react-i18next`

dplaton: You're right, I'll take care of it.

herzog31: Would still be nice to separate the dependencies accordingly. So we know which dependencies are actually needed during "runtime" and which ones are needed during "build time" only.

I think `peerDependencies` are a bonus for when we want to define the dependencies in Venia.

dplaton: > I think peerDependencies are a bonus for when we want to define the dependencies in Venia.

They are used to signal the consumers (not only Venia) that they absolutely need those dependencies to build their app. Also, Webpack doesn't bundle `peerDependencies`, we won't be enforcing our version of (let's say) React on consumers.

herzog31 (approved):

PR #460, opened 2021-01-07
CIF-1787 - Implement ComponentExporter for CIF teaser component (the …

dplaton:
## Description

* Add the required annotations to the CommerceTeaserImpl Sling Model
* Update the action items to include the `categoryId` / `productSlug` to be used by the SPA

## Related Issue

CIF-1787

## Motivation and Context

Export the Sling Model data as JSON 

## How Has This Been Tested?

Unit tests and functional testing

Steps for functional testing:

1/ Install AEM Venia and this branch of the CIF Core Components project
2/ Open the home page in the editor and add a Teaser component, 
3/ Configure the Teaser component as follows:
* Add an asset
* Add a title and a description
* Add a call to action pointing to a category
* Add a CTA pointing to a product
4/ Open the model JSON using the model selector, i.e. `http://localhost:4502/content/venia/us/en.model.json`
5/ Look for the teaser component and check the following:
* The `actions` JSON snippet should be an array
* An `action` should at least have the title and the `identifier` object. For example, for a category, an action could look like this
```JSON
{
  "title": "Shop now",
  "url": "/content/venia/us/en/products/category-page.3.html",
  "entityIdentifier": {
    "entityType": "CATEGORY",
    "value": "3",
    "type": "ID"
  }
}
``` 
* Check that the following properties are exported: title, description, pretitle, asset path.
* Check that the resource type is `core/cif/components/content/teaser/v1/teaser`


herzog31 (approved):
Looks good. Would it make sense to add an integration test to test the actual JSON output of a component instance?

dplaton:
> Looks good. Would it make sense to add an integration test to test the actual JSON output of a component instance?

Yes, it would. I'll do this.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/teaser/CommerceTeaserActionItemImpl.java
```diff
@@ -15,19 +15,28 @@
...
+    private String categoryId;
+    private String productSlug;
 
-    public CommerceTeaserActionItem(String title, String url) {
+    public CommerceTeaserActionItemImpl(String title, String url) {
         this.title = title;
         this.url = url;
     }
 
+    public CommerceTeaserActionItemImpl(String title, String url, String categoryId, String productSlug) {
```
cjelger, line 34: I'm a bit concerned that we hardcode the identifier types, that is, `id` for category and `slug` for product. I think the implementation in `CommerceTeaserImpl` was always pretty much "hardcoding" the fact that we use ids and slugs, but we should maybe avoid hardcoding this any further. Nevertheless, if we want to keep these ids, we should make sure we define some "generic" names that we can use in all components. We should avoid that one component for example exports `productSlug` while another exports `url_key`, etc, maybe we should discuss if we want to define general identifier names or reuse the magento field names.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/teaser/CommerceTeaserActionItem.java
```diff
@@ -0,0 +1,34 @@
...
+import com.adobe.cq.wcm.core.components.models.ListItem;
+
+public interface CommerceTeaserActionItem extends ListItem {
+
+    /**
+     * Returns the category id associated with this teaser action.
+     * 
+     * @return a String representing the category or <code>null</code> if no category id is configured
+     */
+    default String getCategoryId() {
```
cjelger, line 21: I thought we agreed a long time ago that we don't want to use `default` interface methods ... although I think we have a few ones being used. I think we had some issues for that.

dplaton: Ok, I forgot about that. I took the implementation from WCM Core Components as a model

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/teaser/CommerceTeaserActionItemImpl.java
dplaton: I understand your concern - if tomorrow we decide that the category picker should return a different field than `categoryId`, then this parameter name becomes invalid. The best we can do today is replace the names of the methods to be something generic like `getCategoryIdentifier()` and `getProductIdentifier()` 

> We should avoid that one component for example exports productSlug while another exports url_key

But these two represent two different things from the backend. In this case, the component exports both, depending on what is configured - a slug for the product or an URL key for the category. We chose these two because it's easier to work with them. Otherwise, we could've just used "entity id" and do the extra query to get the actual data that we need.

> we should discuss if we want to define general identifier names or reuse the Magento field names.

What would we use instead? We currently identify products by "SKU" and by "slug" (not sure if we use the product id anywhere). We can rename these to something else, but I don't see any point in doing it. The same for categories - we use the id and the url_key depending on what situation we're in.

cjelger: I think you're slightly confusing some fields: what we call "slug" is the product `url_key`, for categories we can use the `url_path` in the URL but only for SEO, not to lookup the category. So for products `slug = url_key`.

In `UrlProvider` we have enums for the id types, so maybe we could rather export something like:
```
productIdentifier: {
  type: SKU | URL_KEY,
  value: xyz
}
```
but it could be something that we use for all components, so we have a consistent way of consuming this in all client-side components. I don't want to be a pain, but it would be nice if we first "design" the export types in a consistent way, to avoid that client-side code has to change or stops working if, for example, a component is configured differently to use skus or slugs.

dplaton: > so maybe we could rather export something like [...]

So the client-side component would use different queries (or rather, different filters for the query) depending on what type of "identifier" it gets from AEM, did I get this correct? 

>to avoid that client-side code has to change or stops working if, for example, a component is configured differently to use SKUs or slugs.

This is a bit unclear to me - are you talking about _the same_ component, i.e. if someone updates the product field in the Teaser component to return an SKU instead of a slug?

cjelger: >So the client-side component would use different queries (or rather, different filters for the query) depending on what type of "identifier" it gets from AEM, did I get this correct?

Yes this is what we already do server-side for the product component at https://github.com/adobe/aem-core-cif-components/blob/master/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/ProductImpl.java#L139

> This is a bit unclear to me - are you talking about the same component, i.e. if someone updates the product field in the Teaser component to return an SKU instead of a slug?

Yes exactly. One problem is that the picker can be changed to set the SKU or SLUG, but most components do not support this (except the product component). I think usually they expect the SKU, but as you see in your PR, the commerce teaser uses the SLUG, so that's why I think we should make sure we don't introduce another level of technical debt where client-side code has "hardcoded" identifier types.

To me this is really something we didn't do well (in the picker and components), so we should probably invest a bit of time in designing this properly to avoid introducing more technical debt in this area.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/common/CommerceIdentifier.java
```diff
@@ -0,0 +1,52 @@
...
+ * An identifier for any commerce entity (category or product). A commerce entity can be identified by more than one key, for example a
+ * product can be identified by the URL key or by the SKU.
+ */
+public interface CommerceIdentifier {
+
+    /**
+     * The type of this identifier
+     */
+    enum IdentifierType {
+        SLUG, URL_KEY, ID
```
cjelger, line 22: `SLUG` is basically `URL_KEY` for products, was this meant to be `SKU, URL_KEY, ID`? We'd have to decide if we use `SLUG` or `URL_KEY` : I'd go for `URL_KEY` because this is already what we use in `UrlProvider`.

dplaton: It was `SKU`, sorry. I'll update it

bundles/core/src/test/resources/test-queries/graphql-requests.log
```diff
@@ -1,46 +1,2 @@
-{__type(name:"ProductAttributeFilterInput"){name,description,inputFields{name,type{name}}}}
```
cjelger, line 1: Sorry I forgot, this file should not be changed like that, please revert

cjelger (approved):

PR #469, opened 2021-01-20
Add user data in dataLayer

laurentiumagureanu:
## Description

User info will be pushed to Adobe Data Layer on every page load

## Related Issue

CIF-1736

## Motivation and Context

Initial dataLayer implementation for CIF included this request

## How Has This Been Tested?

Manually

## Types of changes

## Checklist:


mhaack (approved):

react-components/src/components/AuthBar/authBar.js
```diff
@@ -28,6 +29,15 @@ const AuthBar = ({ showMyAccount, showSignIn }) => {
             getUserDetails();
         }
     }, [getUserDetails]);
+
+    useEffect(() => {
+        if (!isSignedIn) {
+            dataLayerUtils.pushData({ user: 'anonymous' });
```
cjelger, line 35: When not logged in, is it better to set user `anonymous` or not set anything at all? If `anonynous` is not recognised in other systems that this means "not logged in", some other systems might wrongly recognise this is a logged-in user called `anonymous`. I know this is picky, but we don't know how Analytics or Unified Profile would process user `anonymous`.

laurentiumagureanu: Right. AFAIK, Analytics does not consume the dataLayer (yet). It relies on global JS variables (eVar).

The Unified Profile has no direct integration with a page. the data needs to be "stitched" somehow. so it all goes down on how does the JS script that consumes this interprets the data. 

Identifying a user is made in multiple ways. one of them is the login info. another (more reliable) way is a tracking cookie (ECID in the Adobe ecosystem) which is managed by Analytics ( I think ). There is a JS API for accessing that value. Maybe we'll include it as well in the future.

The `user` info has 2 possible datatypes: `String|Object`.
The consuming scripts can check this.
The string value is used to indicate the anonymous user
The object value will contain logged in user info.

I would still add this for an anonymous user. Maybe use an empty string. WDYT?

cjelger: > The user info has 2 possible datatypes: String|Object.
Where is this coming from? if there is some kind of "standard/doc", then it should be clear how this should be set/used?

laurentiumagureanu: Not really but I think this should be put in place at some point. I didn't find any XDM schema that contains the fields needed for the user info.

I think we should just use our own structure and document it so anyone using the dataLayer knows what to expect.

I can drop the `String` datatype and just use `null` when the user is not logged in

cjelger: I let you decide, you have a better overview of the datalayer, I guess it's also important to think about what happens if the user is first not logged in, then logs in, then logs out, and what state changes we want to capture in this case. Maybe it will be more clear if `user` should be set to `anonymous` (my lowest personal preference), `null`, or not set at all.

herzog31 (approved):

alkim91 (approved):

PR #493, opened 2021-02-23
By adding a URL Delegator, implementors/users of the CIF library can customize how product page URLs work in their environment.

froggin:
## By adding a URL Delegator, implementors/users of the CIF library can customize how product page URLs work in their environment.

Users of the CIF library can now have more complete control over how product pages or category pages respond. A `UrlDelegator` gets a list of all available `UrlProvider`s, and then decides which to use based on their `shouldProcess` method (and also based on ranking in the service list). This way, custom client-specific code can be used to implement the `UrlProvider` class and the `UrlDelegator` will be able to use the custom code for processing.

## Motivation and Context

The motivation comes from a client who needed the CIF library but had an existing URL structure for products, so using selectors for different products was not an option. Different workarounds were used for that case, but this is the longer-term thought out solution to allow others to implement their own ways of telling the CIF library what kind of page is a "product" page or not.

## How Has This Been Tested?

This has been tested locally and in an environment with a dispatcher. The code should not affect many other areas as it will default to using the included `UrlProvider` within the existing CIF codebase. The change simply allows someone else to make their own implementation of this, which I have also tested.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change) -- I'm not 100% sure if this is a breaking change but checked this box just out of caution.

## Checklist:


mhaack:
@froggin thanks for the contribution. The `UrlProviders` already allows to have flexible PDP & PLP urls. In combination with Sling mapping configuration and maybe HTTPD rewrites you are very flexible alreads. Also see here: https://experienceleague.adobe.com/docs/experience-manager-cloud-service/commerce/configure/advanced-url-configuration.html?lang=en#product

What would be different with your approach? Can you share some example use cases?

froggin:
@mhaack, what this allows is more than just URL structures (although it can be nice to not have to deal with Sling mapping and HTTPD rewrites sometimes). The specific benefit I believe this offers the most is that product pages can be specifically authored rather than being dependent on a generic PDP template. 

For example, product components can be allowed on any page, and based on the business rules, it could include a product component but not have that product name in the url at all. The logic is controllable per client, per implementation, so control over what product or product list gets returned when becomes more specific.

dplaton:
> product components can be allowed on any page, and based on the business rules, it could include a product component but not have that product name in the url at all. 

@froggin https://github.com/adobe/aem-core-cif-components/pull/505 addresses this use case.

cjelger:
@froggin Thank you for your contribution. We will assign this to one of us in order to carefully test and review the feature. Just one question: what didn't you simply write your own `UrlProvider` and set the right service level so that it replaces our default implementation? This is briefly documented at the bottom of https://github.com/adobe/aem-core-cif-components/wiki/configuration#configuring-and-customising-product-and-category-pages-urls - I'm basically trying to understand if this somehow didn't work, and why you really needed this extra feature.

mhaack:
There have been a lot of rework on the UrlProvider since CIF 2.x and it is possible now to have custom URL implementations as well, so I'm closing this PR.

PR #509, opened 2021-03-04
CIF-1964 - Show staged flag in product component

cjelger:
- add "staged" flag to commerce component title section
- upgrade CIF components library GraphQL server to Magento 2.4.2 schema and adapt unit tests

The `css` for this feature is added to https://github.com/adobe/aem-cif-guides-venia/pull/111

## How Has This Been Tested?

Extended unit tests and added tests to CIF components library.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes and the overall coverage did not decrease.
- [x] All unit tests pass on CircleCi.
- [x] I ran all tests locally and they pass.

PR #510, opened 2021-03-08
CIF-1964 - Show staged flag in product component

cjelger:
This PR is an alternative to https://github.com/adobe/aem-core-cif-components/pull/509 where we introduce a version 2 of the product component so that version 1 can continue working with Magento CE.

If we decide to introduce that `v2` product component, the Venia PR https://github.com/adobe/aem-cif-guides-venia/pull/111 will be modified to update the Venia product proxy component to use that `v2`. 

## How Has This Been Tested?

Extended unit tests and manually tested with Venia.

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v2/product/ProductImpl.java
```diff
@@ -0,0 +1,60 @@
...
+    cache = true)
+public class ProductImpl extends com.adobe.cq.commerce.core.components.internal.models.v1.product.ProductImpl implements Product {
+
+    public static final String RESOURCE_TYPE = "core/cif/components/commerce/product/v2/product";
+
+    @PostConstruct
+    protected void initModel() {
+        super.initModel();
+        if (productRetriever != null) {
+            productRetriever.extendProductQueryWith(p -> p.staged());
```
herzog31, line 40: In case someone wants to customize the component via Sling model delegation and uses the `extendProductQueryWith` or `extendVariantQueryWith` methods and does not include the `staged` attribute, the staging functionality will no longer work. The way these methods work is that the last call wins. Any customizations done before will no longer be respected.

So I would rather:
* either change the way `extendProductQueryWith` and `extendVariantQueryWith` work so that you can call it multiple times and customizations are "merged".
* or create a separate `ProductRetriever` implementation (based on `AbstractProductRetriever`) and add the staged attribute there. I prefer this way, since we will need much more attributes for gift card support.

cjelger: Good point, I'll fix that.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/product/Product.java
```diff
@@ -72,6 +72,17 @@
 
     Boolean isBundleProduct();
 
+    /**
+     * The version 1 of the product component always returns <code>false</code> as it does not support this feature.
+     * The version 2 of the product component does support this feature but it requires a Magento EE instance with
+     * at least version 2.4.2.
+     * 
+     * @return <code>true</code> if the product data contains staged changes, <code>false</code> otherwise.
```
mhaack, line 80: We didn't do this from the beginning, but this is a good chance to introduce `* @since com.adobe.cq.commerce.core.components.models.product 3.1.0` annotations, like WCM core components have as well.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v2/product/ProductImpl.java
cjelger: I finally went for option 1) see the change in `AbstractProductRetriever`

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/product/Product.java
cjelger: Done

dplaton (approved):

mhaack (approved):

herzog31 (approved):

ankitambekar08:
Hi all,

we have created new custom attributes in magento 2.

when we  try to fetch the attributes on aem cif product page component, we are getting below errors
1. **caused by: java.lang.RuntimeException: java.lang.NullPointerException****
2.  Traversal query (query without index

if we try to fetch the same attribute on productteaser component. we are able to see the value.

Aem version : 6.5
Cif connector version : 1.8
magento version : 2.4.

herzog31:
@ankitambekar08 please create an issue via https://github.com/adobe/aem-core-cif-components/issues. Unrelated pull requests are not the right place to discuss this.

PR #518, opened 2021-03-17
CIF-1996: Add copyright headers to CSS and graphql files

laurentiumagureanu:
## Description

Added copyright headers to CSS and graphql files

## Related Issue

CIF-1996

## Motivation and Context

According to a comment on a completely different issue on OSS Advisory Board we should have copyright headers on CSS and GraphQL files as well.

## How Has This Been Tested?

Unit tests
Manually

## Types of changes

## Checklist:


dplaton:
I remember we had some tooling that collected the queries made by the React components (@herzog31 knows more). I think this tooling wouldn't work with the change you did to the `.graphql` files.

react-components/src/queries/__test__/validateQueries.test.js
```diff
@@ -33,7 +33,9 @@ describe.each([['2.4.0', magentoSchema240]])(
 
         it.each(files)('validates the GraphQL request from %s', file => {
             let query = fs.readFileSync(path.join(__dirname, '..', file), 'UTF-8'); // eslint-disable-line
```
cjelger, line 35: You should rather simply dynamically `import` each query/file here as any normal exported js module value.

cjelger (commented):
Related to @dplaton's comment, this will break https://github.com/adobe/commerce-cif-graphql-integration-reference/blob/master/src/documentation/generate.js#L69-L73 but it would also be simple to fix it there.

react-components/src/queries/__test__/validateQueries.test.js
laurentiumagureanu: tried that. didn't work for some reason.

laurentiumagureanu: there seems to be a difference between the parsed query and the imported one:
parsed query:
```  
    {
        kind: 'Document',
        definitions: [
          {
            kind: 'OperationDefinition',
            operation: 'query',
            name: [Object],
            variableDefinitions: [],
            directives: [],
            selectionSet: [Object],
            loc: [Object]
          }
        ],
        loc: { start: 0, end: 146 }
     } 
```
imported query : 
```
    {
        default: {
          kind: 'Document',
          definitions: [ [Object] ],
          loc: { start: 0, end: 147 }
        }
    }
```

laurentiumagureanu: Ignore my comments. I'll update the code.

laurentiumagureanu:
@dplaton, @cjelger I opened a PR for the tool you mentioned 
https://github.com/adobe/commerce-cif-graphql-integration-reference/pull/40

herzog31:
Did anyone ask for an exception? Changing all the graphql files including the webpack build and tests, just for the sake of adding copyright headers is ridiculous. Can't we simply put a license file in the folder where all queries are and say that it applies to all?

I'm fine with the changes to the css files.

PR #525, opened 2021-03-25
@releng - update version numbers after the release of version 1.9.0

LSantha:


README.md
```diff
@@ -44,7 +44,9 @@ The latest version of the AEM CIF Core Components, require the below minimum sys
 
 | CIF Core Components | AEM as a Cloud Service | AEM 6.5 | AEM 6.4 | Magento | Java  |
 | ------------------- | ---------------------- | ------- | ------- | ------- | ----- |
-| 1.8.0               | Continual              | 6.5.7   | 6.4.4.0 | 2.4.0   | 8, 11 |
+| 1.9.0               | Continual              | 6.5.7   | 6.4.4.0 | 2.4.0   | 8, 11 |
+
+Version 1.9.0 is also compatible with Magento 2.4.1 and 2.4.2 .
```
mhaack, line 49: In the past we just added the lastet version in the table.

LSantha: Above the table it writes minimal required versions which would mean that v. 1.9.0 is not compatible with anything older than Magento 2.4.2.

mhaack: Maybe the lets use `2.4.0 - 2.4.2` similar like we do for Java column. So it is clear on the first sight.

LSantha: OK, done.

herzog31 (approved):

mhaack (approved):

PR #534, opened 2021-04-01
CIF-1971: XF placeholder - use UID for category queries

laurentiumagureanu:
## Description

XF for categories uses configured identifier

## Related Issue

CIF-1971

## Motivation and Context

Adopt Magento GraphQL schema changes - Category UID

## How Has This Been Tested?

Unit test

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/experiencefragment/CommerceExperienceFragmentImpl.java
```diff
@@ -154,18 +155,20 @@ private String buildQueryForProduct(String sku) {
     private String getQueryForCategory() {
         // Parse category identifier in URL
         Pair<CategoryIdentifierType, String> identifier = urlProvider.getCategoryIdentifier(request);
-        String id = null;
+        String categoriesIdentifier = null;
 
-        if (CategoryIdentifierType.ID.equals(identifier.getLeft())) {
-            id = identifier.getRight();
+        List<CategoryIdentifierType> validCategoryIdentifiers = Arrays.asList(CategoryIdentifierType.ID, CategoryIdentifierType.UID);
```
herzog31, line 160: This could be immutable and static :)

herzog31 (approved):
Looks good to me.

I think we need to make the behaviour clear in the migration documentation, that XFs need to be tagged with category identifiers that match those in the URL based on the URLProvider configuration.

* URLProvider configured for UIDs -> only XFs with UID reference will be considered
* URLProvider configured for IDs -> only XFs with ID reference will be considered

dplaton (approved):

PR #538, opened 2021-04-07
CIF-2022: Fix category URL for UID

laurentiumagureanu:
## Description

Adds support for generating category URLs when UID support is enabled or UID is used as selection for custom category pages

## Related Issue

CIF-2022

## Motivation and Context

Adopt Magento GraphQL schema changes - Category UID

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/categorylist/CategoriesRetriever.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/categorylist/CategoriesRetriever.java
@@ -19,9 +19,15 @@ import com.adobe.cq.commerce.core.components.services.UrlProvider;
 import com.adobe.cq.commerce.magento.graphql.CategoryTreeQueryDefinition;
 
 public class CategoriesRetriever extends AbstractCategoriesRetriever {
+    private boolean enableUIDSupport;
 
     CategoriesRetriever(MagentoGraphqlClient client) {
         super(client);
+    }
+
+    CategoriesRetriever(MagentoGraphqlClient client, boolean enableUIDSupport) {
+        super(client);
+        this.enableUIDSupport = enableUIDSupport;
     }
 
     @Override
@@ -33,7 +39,7 @@ public class CategoriesRetriever extends AbstractCategoriesRetriever {
                 .position()
                 .image();
 
-            if (identifierType == UrlProvider.CategoryIdentifierType.UID) {
+            if (enableUIDSupport || identifierType == UrlProvider.CategoryIdentifierType.UID) {
                 q.uid();
             }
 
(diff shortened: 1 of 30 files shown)


herzog31 (approved):
Do we have some kind of documentation of migration guide, when a customer should start using `uidAndUrlPath`?

laurentiumagureanu:
@herzog31 CIF-2021 should cover the docs

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/categorylist/CategoriesRetriever.java
```diff
@@ -32,6 +39,10 @@ protected CategoryTreeQueryDefinition generateCategoryQuery() {
                 .position()
                 .image();
 
+            if (enableUIDSupport || identifierType == UrlProvider.CategoryIdentifierType.UID) {
```
mhaack, line 42: We should also only query `id` if we are using it. See line 36.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/categorylist/FeaturedCategoryListImpl.java
```diff
@@ -164,10 +172,15 @@ private void initModel() {
...
         List<CategoryTree> categories = categoriesRetriever.fetchCategories();
         for (CategoryTree category : categories) {
-            Map<String, String> params = new ParamsBuilder()
+            ParamsBuilder paramsBuilder = new ParamsBuilder()
                 .id(category.getId().toString())
-                .urlPath(category.getUrlPath())
-                .map();
+                .urlPath(category.getUrlPath());
+
+            if (enableUIDSupport || categoryIdentifierType == CategoryIdentifierType.UID) {
```
mhaack, line 179: Same as above, if `uid` is used we should not add `id` to the params.

herzog31: At this point we don't know. It depends on the configuration of the URLProvider.

mhaack: Arg, yes you are right we must query all options used by the URLProvider.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/categorylist/CategoriesRetriever.java
herzog31: The condition here is for if it's supported to query a category by UID. So you can query for the UID anytime, no condition needed.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/categorylist/FeaturedCategoryListImpl.java
mhaack: Maybe the URLProvider should provide a method to return how the configuration is and what attribute is needed, this can then be used by the query.

PR #539, opened 2021-04-08
CIF-1785 - Implement ComponentExporter for category carousel component

cjelger:
- fixed returned resource type for json export for all components

The `categorycarousel` reuses the `FeaturedCategoryList` sling model but the model wrongly returned the "hard-coded" CIF component type. To align with what the WCM core components do, and what is done in the datalayer, I changed all components already implementing `ComponentExporter` to return the actual resource type. This "fixes" the export for the `categorycarousel` component and for all other components.

## How Has This Been Tested?

Adapted unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/categorylist/FeaturedCategoryListImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/categorylist/FeaturedCategoryListImpl.java
@@ -240,6 +240,6 @@ public class FeaturedCategoryListImpl extends DataLayerComponent implements Feat
 
     @Override
     public String getExportedType() {
-        return RESOURCE_TYPE;
+        return resource.getResourceType();
     }
 }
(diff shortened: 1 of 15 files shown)


laurentiumagureanu (approved):

dplaton:
> the model wrongly returned the "hard-coded" CIF component type

This was by design. In the SPA you need to map the React component to the AEM resource type using a call like
```javascript
MapTo(`aem/resource/type`)(ReactComponent)
```
If you use the "Venia" resource type then you are creating a hard link between a project name and the React Components, greatly reducing SPA components reusability, since all the component mappings would have to be adjusted for every project.

cjelger:
@dplaton Thanks for clarifying this, but I checked the WCM Core components and they export the actual/real resource type with `resource.getResourceType();` ... I actually think that the archetype would (should?) generate the correct SPA and React components with the proper `MapTo(...)` values, WDYT? I'll also check this.

dplaton:
@cjelger that's an interesting approach, but I think we have a slightly different workflow: our React components (at least the ones that currently consume these Sling Models are meant to work as a PWA Studio extension. Developers would just have to install this extension as a dependency to their PWA Studio app, configure it, and that's it. There's no archetype involved here.

I must admit the situation is a bit murky here. We don't have support from the editor (yet) which is the main pain point, that's why we didn't bother much with the rest of it (like how a project would be bootstrapped etc)

dplaton:
There are solid reasons for this update, but this will break some existing code. Since the SPA Editor feature is still WIP I'll put this into "draft" status for the moment.

dplaton:
Closing for now, but I'll leave the branch in the repository.

PR #551, opened 2021-04-22
@trivial - update readme after release

mhaack:


LSantha:
@mhaack , are there any ee features in this release? I thought that we only require Magento 2.4.2.

mhaack:
I used 2.4.2ee because of https://github.com/adobe/aem-core-cif-components/blob/master/parent/pom.xml#L109, we are not using specific ee features.

LSantha:
OK, that's only the ee version of the library but not Magento. Based on the readme update the users will think that they need Magento ee already.

mhaack:
Updated the readme, removed the ee

LSantha (approved):

herzog31 (approved):

PR #555, opened 2021-04-27
CIF-2049 - implement UID support for button component and use URLProvider for the component

### LINKED ISSUE #553: Button component doesn't use UrlProvider to create product and category links
### Expected Behaviour
When using the [CIF Button (v1) component](https://github.com/adobe/aem-core-cif-components/tree/master/ui.apps/src/main/content/jcr_root/apps/core/cif/components/content/button/v1/button), I expected it to create URLs that are formatted based on what's defined in the UrlProvider OSGi config.

Expected: /content/luma/language-masters/en/category/category-page.21.html/women/tops-women

### Actual Behaviour

The behavior that I'm seeing is that it's missing the suffix and therefore breaks the special sling mapping that's set up.

Actual: /content/luma/language-masters/en/category/category-page.21.html

#### Steps to Reproduce

Drop the CIF Button on a page and link to a category from the authoring dialog.

#### Platform and Version
AEM 6.5.6

#### Logs taken while reproducing problem
ButtonImpl:
```
    private String constructUrl(final String pagePath, final String urlKey) {
        return String.format("%s.%s.html", pagePath, urlKey);
    }
```

mhaack:
## Description

Fixes the Button Component implementation which was very outdated.
 - add UID support 
 - use URLProvider to generate PDP/PLP links

## Related Issue

Fixes #553

CIF-2049

## How Has This Been Tested?

Extended existing unit tests and added extra unit tests.

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/button/CategoryRetriever.java
```diff
@@ -0,0 +1,48 @@
...
+    CategoryRetriever(MagentoGraphqlClient client, boolean enableUIDSupport) {
+        super(client);
+        this.enableUIDSupport = enableUIDSupport;
+    }
+
+    @Override
+    protected CategoryTreeQueryDefinition generateCategoryQuery() {
+        CategoryTreeQueryDefinition categoryTreeQueryDefinition = q -> {
+            q.id().urlPath();
+            if (enableUIDSupport || categoryIdentifierType == UrlProvider.CategoryIdentifierType.UID) {
```
herzog31, line 37: You don't need this condition here. You can always query for it, since it's available with 2.4.2 and that's the min. required version.

herzog31 (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/button/CategoryRetriever.java
mhaack: Change it, code now always queries for uid and use it in the button component impl. This also simplifies the code a lot since `enableUIDSupport` is not needed any more.

PR #556, opened 2021-04-28
CIF-2032: url_path support for category pages

laurentiumagureanu:
## Description

Support using url_path for category pages.

## Related Issue

CIF-2032

## Motivation and Context

Support using url_path for category pages.

## How Has This Been Tested?

Unit test

## Types of changes

## Checklist:


mhaack:
Can we also add a corresponding sample /etc/map which works with the url_path?

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/retriever/AbstractCategoryRetriever.java
```diff
@@ -260,12 +260,21 @@ public String generateQuery(String identifier) {
      * @return GraphQL query as string
      */
     public Pair<CategoryListArgumentsDefinition, CategoryTreeQueryDefinition> generateCategoryQueryArgs(String identifier) {
-        FilterEqualTypeInput identifierFilter = new FilterEqualTypeInput().setEq(identifier);
+        FilterEqualTypeInput identifierFilter = new FilterEqualTypeInput().setEq(identifier.replaceAll("_", "/"));
```
Buuhuu, line 263: You are missing this `replaceAll` in the `BreadcrumbRetriever.java`. 

Is there any way to keep the logic in a single place?

bundles/core/src/main/java/com/adobe/cq/commerce/core/search/internal/services/SearchResultsServiceImpl.java
```diff
@@ -141,20 +143,34 @@ public SearchResultsSet performSearch(
...
+            String categoryQueryString = generateCategoryQueryString(categoryRetriever).get();
+            LOGGER.debug("Generated category query string {}", categoryQueryString);
+            GraphqlResponse<Query, Error> categoryResponse = magentoGraphqlClient.execute(categoryQueryString);
+            Query categoryData = categoryResponse.getData();
+
+            if (categoryData != null) {
+                List<CategoryTree> categories = categoryData.getCategoryList();
+                if (CollectionUtils.isNotEmpty(categories)) {
+                    category = categories.get(0);
+                    ((SearchOptionsImpl) searchOptions).setCategoryId(category.getId().toString());
```
Buuhuu, line 158: The SearchResultsService is public API, hard casting here implicitly introduces a dependency to an un-exported package. (the one containing the SearchOptionsImpl).

I would recommend to add a constructor `SearchOptionsImpl(SearchOptions)`. It creates a 1:1 copy of the given SearchOptions as `SearchOptionsImpl`, which you can then mutate. 

Not to handle in this PR, but SearchOptions should be DTO w/o an interface and the logic should be kept in a service implementation.

Buuhuu (changes requested):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/retriever/AbstractCategoryRetriever.java
Buuhuu, line 263: Can any part of the identifier contain underscores? If so we have to consider escaping. 

Actually I would recommend to use dashes instead of underscores.

https://developers.google.com/search/docs/advanced/guidelines/url-structure?hl=en&visit_id=637568438766532832-2808360326&rd=1

> Consider using punctuation in your URLs. The URL http://www.example.com/green-dress.html is much more useful to us than http://www.example.com/greendress.html. We recommend that you use hyphens - instead of underscores _ in your URLs.

mhaack: The url format and with that at least the `url_key` and `url_path` are controlled by the commerce backend. For Magento https://docs.magento.com/user-guide/catalog/catalog-urls.html, other solutions might have different formats.

So while we technically can totally change this I think we still want to align with what is configured for the backend.

Buuhuu: You mean, for Magento parts of the identifier (=url_key) may contain dashes right? If so we cannot use dashes for Mangeto but as you said others may do that differently. May be the char to replace the path separator needs to be configureable? 

However do we distinguish between author and publish? For publish a suffix should work better than a selector and there we can have slashes. If not mixing underscores in the url is not a good idea because Google joins words on them while splitting on dash. SEO is irrelevant for author.

mhaack: >Maybe the char to replace the path separator needs to be configurable?
Yes good idea, however we want to make the entire URL handling simpler. So config with default is ok to me.

>For publish a suffix should work better than a selector and there we can have slashes.
We use selectors because of caching behavior on the AEM dispatcher. Plus there is typically some URL rewriting setup on the publish as well. The public visible URL will typically not use any selector or suffix.

Buuhuu: Suffixes can as well be cached on dispatcher:

/path/to/category-page.html => editing purpose, 400 Bad Request on publish
/path/to/category-page.cat.html/url_path.html

Selectors' main purpose is view selection (they are taken into account for script resolution) and not data transport. I agree that using suffix on author is tricky in the editor but they are much more suited to transfer information (esp. hierarchical information)

Handling the rewriting/SEO handling of urls on dispatcher level spreads logic across systems which makes the system harder to maintain as a whole. (Think about e2e tests that would require a dispatcher to exists).

mhaack: > Suffixes can as well be cached on dispatcher

I know but requires extra config if I'm right?

> Handling the rewriting/SEO handling of urls on dispatcher level spreads logic across systems which makes the system harder to maintain as a whole. (Think about e2e tests that would require a dispatcher to exists).

I totally agree here, however it is still a common practice to use /etc/map and mod_rewrite.

laurentiumagureanu: Could be done in `AbstractRetriever` class but that class is general purpose and should not contain methods specific to categories or products

laurentiumagureanu: I'll use ~ instead of _
Looks like Magento doesn't allow ~ in URL key definitions so we are safe

Buuhuu: According to RFC1738 Section 2.2 is a unsafe character and we must encode it. Let's use with the underscore for now.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/retriever/AbstractCategoryRetriever.java
```diff
@@ -81,6 +81,28 @@
...
+     * Generates the filter for a category or categoryList query
+     *
+     * @param identifiersFilter The filter portion to be used by CategoryFilterInput.
+     * @param categoryIdentifierType The category identifiers used byt the identifiersFilter
+     */
+    public static CategoryFilterInput generateCategoryFilter(FilterEqualTypeInput identifiersFilter,
+        UrlProvider.CategoryIdentifierType categoryIdentifierType) {
+        CategoryFilterInput filter;
+
+        if (UrlProvider.CategoryIdentifierType.UID.equals(categoryIdentifierType)) {
```
Buuhuu, line 94: Am I right that you are missing the `URL_PATH` here. The previous, changed code was a switch

```
switch (categoryIdentifierType) {
            case ID:
                filter = new CategoryFilterInput().setIds(identifierFilter);
                break;
            case UID:
                filter = new CategoryFilterInput().setCategoryUid(identifierFilter);
                break;
            case URL_PATH:
                filter = new CategoryFilterInput().setUrlPath(identifierFilter);
                break;
            default:
                throw new RuntimeException("Category identifier type is not supported");
        }
```

bundles/core/src/main/java/com/adobe/cq/commerce/core/search/internal/models/SearchOptionsImpl.java
```diff
@@ -50,6 +50,24 @@ public SearchOptionsImpl() {
         attributeFilters = new HashMap<>();
     }
 
+    public SearchOptionsImpl(SearchOptions searchOptions) {
+        attributeFilters = searchOptions.getAttributeFilters();
```
Buuhuu, line 54: This should probably be `SearchOptions#getAllFilters()`? 

For the implementation of `getAttributeFilters()` the returned map does not contain the category specific fields.

Buuhuu (changes requested):

bundles/core/src/main/java/com/adobe/cq/commerce/core/search/internal/models/SearchOptionsImpl.java
laurentiumagureanu: fixed in the lines below

Buuhuu (approved):

PR #565, opened 2021-05-11
CIF-2034: XF placeholder - make it independent from URL selector

laurentiumagureanu:
## Description

XFs are to be tagged with the category identifier (ID or UID) as per CA config.
If the URLProvider is configured to use another category field, then a graphQL query is executed to fetch the right identifier.
Same applies for products

## Related Issue

CIF-2034

## Motivation and Context

The current XF placeholder component is limited in finding related XF for product or category pages based on the URL selector only. This requires the URL selector used to be the same the XFs are tagged with.

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/experiencefragment/CommerceExperienceFragmentImpl.java
```diff
@@ -95,9 +101,30 @@
 
     private Resource xfResource;
     private String name;
+    private AbstractCategoryRetriever categoryRetriever;
```
herzog31, line 104: Normally the retrievers should be exposed via getters, so they can be accessed when applying Sling model delegation. But I don't think it's needed here, since the retrievers are only required for identifier lookup and I don't see a use case why the query would be extended.

herzog31 (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/experiencefragment/CommerceExperienceFragmentImpl.java
```diff
@@ -160,8 +190,21 @@ private String getQueryForCategory() {
         Pair<CategoryIdentifierType, String> identifier = urlProvider.getCategoryIdentifier(request);
         String categoriesIdentifier = null;
 
-        if (VALID_CATEGORY_IDENTIFIERS.contains(identifier.getLeft())) {
-            categoriesIdentifier = identifier.getRight();
+        if (identifier.getRight() != null) {
```
Buuhuu, line 193: `VALID_CATEGORY_IDENTIFIERS` is now unused and may be removed.

PR #582, opened 2021-06-04
CIF-2120: Add StorefrontEvents to PDP

laurentiumagureanu:
## Description

The product component generates the context required by Magento Storefront Events SDK

## Related Issue

CIF-2120

## Motivation and Context

Required by the product recommendations engine

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:


ui.apps/src/main/content/jcr_root/apps/core/cif/clientlibs/storefront-events/js/collector.js
```diff
@@ -17,3 +17,18 @@ import mse from '@adobe/magento-storefront-events-sdk';
 
 // Expose Magento Storefront Events SDK on the global window object
 window.magentoStorefrontEvents = mse;
+
+const processProductStorefrontData = () => {
```
herzog31, line 21: We should wrap this part into an IIFE and also check for the DOM to be ready. Similar to https://github.com/adobe/aem-core-wcm-components/blob/development/content/src/scripts/datalayer/v1/datalayer.js#L100-L104.

herzog31 (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/storefrontcontext/ProductStorefrontContext.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ ******************************************************************************/
+package com.adobe.cq.commerce.core.components.storefrontcontext;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public interface ProductStorefrontContext extends CommerceStorefrontContext {
+
+    @JsonProperty("productId")
+    Integer getId();
```
mhaack, line 21: I assume id being an Integer is given by the events format? Can we make it String also to be future ready with UID and other ID formats?

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/storefrontcontext/ProductStorefrontContextImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private final ProductInterface product;
+
+    public ProductStorefrontContextImpl(ProductInterface product) {
+        this.product = product;
+    }
+
+    @Override
+    public Integer getId() {
+        return 0;
```
mhaack, line 29: ? We ignore the actual product id?

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/storefrontcontext/ProductStorefrontContext.java
herzog31: It has to be an integer according to the Event SDK schema: https://github.com/adobe/magento-storefront-events-sdk/blob/main/src/types/schemas/product.ts#L7.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/storefrontcontext/ProductStorefrontContextImpl.java
herzog31: We have confirmation that internally SKU is used to match products. The id however is still a mandatory field in the schema. So as long as UID support is missing, we decided to just go with a dummy value that fulfils the schema.

mhaack: Ok maybe add a //TODO comment about this behavior and that we work around bad Events SDK design.

mhaack (approved):

PR #603, opened 2021-06-28
CIF-2137: remove deprecated methods from models

Buuhuu:
This change removes methods/fields marked deprecated from all Java classes. 

## Related Issue

CIF-2137

## Motivation and Context

Major release of the CIF Core Components.

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- *    Copyright 2019 Adobe. All rights reserved.
+ *    Copyright 2021 Adobe. All rights reserved.
  *    This file is licensed to you under the Apache License, Version 2.0 (the "License");
  *    you may not use this file except in compliance with the License. You may obtain a copy
  *    of the License at http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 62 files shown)


bundles/core/src/test/java/com/adobe/cq/commerce/core/components/internal/models/v1/button/CategoryRetrieverTest.java
```diff
@@ -80,6 +74,7 @@ public void testCategoryUrlPathQueryFallback() {
 
     @Test
     public void testExtendCategoryQuery() {
+        // TODO: check if that is redundant with CategoryRetrieverTest#testCategoryUrlPathQueryById()
```
mhaack, line 77: Yes I think this is redundant, however with CIF-2062 & CIF-2153 the entire CategoryRetrieverTest.java for button component will not be needed. So ok for now.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/button/ButtonImpl.java
```diff
@@ -84,6 +84,9 @@
     @Via(type = ResourceSuperType.class)
     private Button button;
 
+    @Self
```
mhaack, line 87: For all components `@Self(injectionStrategy = InjectionStrategy.OPTIONAL)` is used but not here? Is that by intention?

Buuhuu: yes because the ButtonImpl's defaultInjectionStrategy is OPTIONAL.

As I don't have a strong opinion on that but prefer more explicit settings, I will change that, thanks.

mhaack: +1 to change to simply have it consistent

mhaack (approved):

herzog31 (approved):

dplaton (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java
```diff
@@ -11,258 +11,36 @@
...
- * This is a wrapper class for {@link GraphqlClient}. The constructor adapts a {@link Resource} to
- * the GraphqlClient class and also looks for the <code>magentoStore</code> property on the resource
- * path in order to set the Magento <code>Store</code> HTTP header. This wrapper also sets the custom
- * Magento Gson deserializer from {@link QueryDeserializer}.
+ * This interface gives access to a {@link GraphqlClient} configured for a given context.
+ * <p>
+ * It is implemented as adapter for {@link SlingHttpServletRequest} and {@link Resource} where the full feature set can only be used with
+ * a {@link SlingHttpServletRequest} as an adaptable.
+ *
+ * @see MagentoGraphqlClientImpl for what feature are supported for any of the adaptables
```
LSantha, line 36: Ideally the interface should be expressive enough to tell what you can expect without looking at the implementation and referring the implementation here could be avoided.

Buuhuu: True, but with Sling Models a lot of things are implicit and I am not sure how to express that adapting from a `Resource` provides only a subset of features than adapting from `SlingHttpServletRequest`.

I think as this is oss and in the same page it is good enough for a user to understand the difference.

LSantha: You can simply add it to the Javadoc of the interface.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/client/MagentoGraphqlClientImpl.java
```diff
@@ -0,0 +1,311 @@
...
+
+    public MagentoGraphqlClientImpl(SlingHttpServletRequest request) {
+        this.request = request;
+        this.resource = request.getResource();
+    }
+
+    /**
+     * @deprecated use sling models in all cases instead
+     */
+    @Deprecated
```
LSantha, line 98: If this PR is about deprecation, I wonder if there is any chance to get rid of this constructor? :-)

Buuhuu: Would be great, yes. But that would require touching a lot of unit tests. I would prefer to keep it. It is now in a private package and removing it will not cause a breaking change. 

We can create a follow up story to remove it and fix the unit tests.

PR #611, opened 2021-07-06
Issue/brand slug

nishneo:
## Description

Brandslug from page properties is not coming on page title when using CIF page component


mhaack:
Brand slug was added in https://github.com/adobe/aem-core-wcm-components/releases/tag/core.wcm.components.reactor-2.14.0 we currently still have only WCM core components 2.9.0 dependency. This should be aligned.

dplaton (approved):

Buuhuu (commented):
@LSantha can this be done together with CIF-2211?

mhaack:
With CIF-2211 closed now can this be merged? @LSantha wdyt?

mhaack:
Replaced by #662

PR #619, opened 2021-07-13
CIF-1926 - Minimal default styling for navigation

LSantha:
Implemented default styling for navigation component.

## Related Issue

CIF-1926

## Screenshot
navigation

## How Has This Been Tested?

Manually.

## Types of changes

## Checklist:


mhaack:
Why is this PR needed at all? We have some styles for the Venia project and the AEM Archetype should (?) have the minimal styles for the navigation anyway?

herzog31 (approved):

LSantha:
This PR is needed because the core WCM navigation, that we extend, doesn't provide any styling out of the box with the component, so the styling you have is what you get from the browser for the markup of the navigation component: the usual rendering of the ul/li structures as a vertical multilevel list. For anything slightly more complex than our toy Venia catalog this provides a quite suboptimal user experience in the newly generated project where you only see this list in the upper part of the page and you have to scroll down every time one or two pages or more to see any actual content of the page. The simple style I added is a remedy for this and it gives you manageable page even with our XL catalog.

mhaack:
> because the core WCM navigation, that we extend, doesn't provide any styling out of the box with the component, so the styling you have is what you get from the browser for the markup of the navigation component

Why would this not be enough for us as well? To me Core Components should not have any styling if possible, so I'm fine to have that in AEM Archetype as starter for a project.

LSantha:
It's basically as I explained above, your pages are full of navigation which is always in your way. If you generate a new project with the archetype and pass the XL catalog endpoint for it you will see what I mean.

IMHO there is a fundamental difference here between  core WCM navigation and commerce navigation component. For a project generated without commerce you only have a start page in the generated project, so your navigation component is really small using little real estate in your page.
When you generate the project with commerce enabled all all your pages we generate with the out of the templates have already the commerce navigation component configured. You can simply pass your existing catalog endpoint with real data to the archetype command so that when you generate and install the project in AEM it already works for the very beginning. This implies that with commerce navigation it may happen that from the very first moment you may have a large navigation structure or simply large enough to be a real pain for you on the pages and something really annoying for the customer. 
So that is what I'm trying to avoid with the minimal default styling for the navigation component.
In any case that less than 30 lines of sparse CSS shouldn't be an issue to any user when they implement the real navigation styling for their site.

I hope that helps.

mhaack (approved):

PR #626, opened 2021-07-16
CIF-2211 - Update core WCM dependency to latest for CIF components

LSantha:
Updated version number to 2.17.2 and fixed test failures.

## Related Issue

CIF-2211

## Motivation and Context

JUnit.

## Types of changes

## Checklist:


.circleci/config.yml
```diff
@@ -72,13 +72,13 @@ executors:
...
         <<: *docker_auth
-      - image: docker-adobe-cif-release.dr-uw2.adobeitc.com/circleci-aem-cloudready:5343-azul
+      - image: docker-adobe-cif-release.dr-uw2.adobeitc.com/circleci-aem-cloudready:5607-azul
         <<: *docker_auth
   test_executor_655:
     docker:
       - image: docker-adobe-cif-release.dr-uw2.adobeitc.com/circleci-qp:6.4.4-azul-jdk11
         <<: *docker_auth
-      - image: docker-adobe-cif-release.dr-uw2.adobeitc.com/circleci-aem:6.5.8-azul
+      - image: docker-adobe-cif-release.dr-uw2.adobeitc.com/circleci-aem:6.5.9-azul
```
Buuhuu, line 81: should we update the readme saying we support 6.5.9+ instead 6.5.7 now?

mhaack: +1

mhaack: Plus we should keep the `uber-jar` version in parent pom.xml in sync as well.

LSantha: In the readme 6.5.7 appears as minimum system requirement.  This PR doesn't have anything that would require AEM 6.5.9, I've updated the CI build only to make sure that we are compatible with the latest AEM releases.
The uber-jar version is still AEM 6.4.4 . 
So, should we enforce AEM 6.5.9 then? (meaning whoever wants to use the CIF components should upgrade to AEM 6.5.9)

herzog31: If we test with 6.5.9, we would never find out if we introduced anything that breaks with 6.5.7.

mhaack: +1 we should test, require and document only one version.

>This PR doesn't have anything that would require AEM 6.5.9

If that is the case we should revert the container version if not really needed.

LSantha:
I've reverted the AEM image to 6.5.8 here and in the Venia PR (https://github.com/adobe/aem-cif-guides-venia/pull/165).
Both are passing.

Buuhuu (approved):

mhaack (approved):

PR #629, opened 2021-07-19
CIF-2197: move UrlFormat into a public package

Buuhuu:
## Description

To provide the `UrlFormat` as SPI to implementors it was moved into a public package together with the `UrlProvider`. Both are in a new package now to achieve distinguished versioning of them.

Implementation details have been moved to the `UrlProviderImpl` and some dead code was eliminated.

This also adds the urlKey to all invocations of `toCategoryUrl(...,params)` in order to provide a consistent set of parameters to the implementations. Furthermore the category uid parameter is now (again) added consistently to be used by UrlFormats.

## Related Issue

CIF-2197
CIF-2062

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/services/UrlProviderImpl.java
```diff
@@ -316,9 +383,12 @@ public String getCategoryIdentifier(SlingHttpServletRequest request) {
             return identifier;
         }
 
-        Map<String, String> categoryIdentifiers = categoryPageUrlFormat.parse(request.getRequestPathInfo());
+        Map<String, String> categoryIdentifiers = categoryPageUrlFormat
+            .parse(request.getRequestPathInfo(), request.getRequestParameterMap());
 
-        if (categoryIdentifiers.containsKey(URL_KEY_PARAM)) {
+        if (categoryIdentifiers.containsKey(UID_PARAM)) {
```
mhaack, line 389: This will re-add support for UIDs in the URL, do we really want this?

Buuhuu: If a customer implements a custom UrlFormat they should have all possible choices. In Magento it makes not much sense to use the uid, but if you implement a custom commerce backend and map it to something human readable it may make sense.

I think it is ok to "re-add" it ... not for the default configuration but the SPI.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/services/urls/UrlFormat.java
```diff
@@ -11,24 +11,47 @@
...
+ * implementation. Implementing a {@link UrlFormat} is optional.
+ */
+@ConsumerType
 public interface UrlFormat {
 
+    /**
+     * The service registration property used to identify the purpose of the {@link UrlFormat}. It can either be set to
+     * {@link UrlFormat#PRODUCT_PAGE_URL_FORMAT} or {@link UrlFormat#CATEGORY_PAGE_URL_FORMAT}.When
+     */
+    String PROP_USE_AS = "useAs";
```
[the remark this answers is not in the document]
Buuhuu: Would make sense, but the service registration does not guarantee type-safety:

```
@Component( property = {  UrlFormat.PROP_USE_AS + "=" + UrlFormat.PRODUCT_PAGE_URL_FORMAT } )
```

```
@Reference(target = "(" + UrlFormat.PROP_USE_AS + "=" + UrlFormat.PRODUCT_PAGE_URL_FORMAT + ")")
```

So it does not make a different other than adding minimal overhead. Les keep it like that.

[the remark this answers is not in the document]
mhaack: Ok.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/services/UrlProviderImpl.java
mhaack: Ok to me. I agree on the 3rd party case. To have products & category in sync we should also support UID option for products.

Buuhuu: Makes sense but the ProductFilterInput does not provide an option to filter products by uid

https://devdocs.magento.com/guides/v2.4/graphql/queries/products.html#ProductFilterInput

So if it is used it may only be used for formatting a url. Also we already have the SKU which allows to have an identifier in the url that does not require a query to retrieve it from the backend.

mhaack: >Makes sense but the ProductFilterInput does not provide an option to filter products by uid

Ah right.

mhaack (approved):

PR #631, opened 2021-07-20
CIF-2223 - Productlist item shows placeholder image and image at the same time

LSantha:
Display placeholder image only if the product has no image.

## Related Issue

CIF-2223

## How Has This Been Tested?

Manually.

## Types of changes

## Checklist:


ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/productcollection/v2/productcollection/item.html
```diff
@@ -20,10 +20,10 @@
         data-sku="${item.SKU}"
         href="${item.URL}">
         <div class="productcollection__item-images">
-            <img class="productcollection__item-image productcollection__item-image--placeholder"
-                alt=""
-                src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAQAAAAFCAQAAADIpIVQAAAADklEQVR42mNkgAJGIhgAALQABsHyMOcAAAAASUVORK5CYII="/>
-            <img class="productcollection__item-image" src="${item.imageURL}" alt="${item.title}"/>
+            <img data-sly-test.hasImage="${item.imageURL}" class="productcollection__item-image" src="${item.imageURL}" alt="${item.title}"/>
+            <img data-sly-test="${!hasImage}" class="productcollection__item-image productcollection__item-image--placeholder"
```
Buuhuu, line 24: we should not use a `` here as this would require a customer to overwrite the entire item.html to change it. If we use a `<div>` and apply the same image as background-image it can be changed using CSS. 

Accessibility-wise it should be fine as well as the placeholder is a decorative image.

LSantha: I have especially checked this and actually it's quite easy to style the current placeholder  element. The background color you saw in Venia is also defined in CSS and you can set a background image for the img element in CSS as well. This is because the inline image data represents a 4x5 sized PNG with all transparent pixels.
I think the reason for using the img element for teh placeholder was to provide similar styling capabilities easily with the real image. If you would use a div here you would need to style it quite differently from the img to have the same effect and also the element - mod CSS classes in use here wouldn't work that well.
So, as far as I can tell, this LGTM.

Buuhuu: Thanks @LSantha, good to know. I thought the 4x5 sized PNG was grey :D

Buuhuu (approved):

mhaack (approved):

PR #639, opened 2021-07-26
CIF-2188 - Expose HTML id attribute in component edit dialogs

LSantha:
* adding field for HTML id attribute in edit dialog of components
 * reusing the HTML id generation of core WCM components
 * updated unit tests

## Related Issue

CIF-2188

## How Has This Been Tested?

JUnit, manually.

## Types of changes

## Checklist:


herzog31 (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/datalayer/DataLayerComponent.java
```diff
@@ -15,23 +15,28 @@
...
 import com.adobe.cq.commerce.core.components.datalayer.CategoryData;
+import com.adobe.cq.wcm.core.components.models.Component;
 import com.adobe.cq.wcm.core.components.models.datalayer.AssetData;
 import com.adobe.cq.wcm.core.components.models.datalayer.ComponentData;
+import com.adobe.cq.wcm.core.components.util.ComponentUtils;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
 public abstract class DataLayerComponent {
-    public static final String ID_SEPARATOR = "-";
+    public static final String ID_SEPARATOR = ComponentUtils.ID_SEPARATOR;
```
mhaack, line 32: Why redefining the const instead of using `ComponentUtils.ID_SEPARATOR` directly via static import?

dplaton (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/datalayer/DataLayerComponent.java
LSantha: The ID_SEPARATOR is used at more other places in the codebe in sub-classes of DataLayerComponnet. If you define it here you don't need any static import in many other files. In fact the constant was defined earlier, I've just aligned the definition with ComponentUtils.

mhaack: I know the constant was defined earlier and it was wrong. So instead of having the same constant twice I would prefer to remove it and only use `ComponentUtils.ID_SEPARATOR`. The static import should not hurt.

Buuhuu: +1 - this is more explicit.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/datalayer/DataLayerComponent.java
```diff
@@ -69,10 +74,14 @@ protected ComponentData getComponentData() {
     }
 
     protected String generateId() {
-        String resourceType = resource.getResourceType();
-        String prefix = StringUtils.substringAfterLast(resourceType, "/");
-        String path = resource.getPath();
-        return StringUtils.join(prefix, ID_SEPARATOR, StringUtils.substring(DigestUtils.sha256Hex(path), 0, 10));
+        if (wcmComponent == null) {
```
Buuhuu, line 77: how can `wcmComponent` be `null`?

We only have 2 models using an optional default injection strategy (button v1 and v2). For all others the `wcmComponent` is required.

LSantha: DataLayerComponent is also extended by classes which are not Sling models, we handle that case here.

Buuhuu: Can you add this detail to the javadoc of the class while you are on it?

LSantha: What detail?

Buuhuu: That this abstract class may be extended by classes that are annotated as sling models and others that are not.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/datalayer/DataLayerComponent.java
LSantha: I've inlined the ID_SEPARATOR constant.

mhaack (approved):

laurentiumagureanu (approved):

PR #641, opened 2021-07-28
CIF-2078: Improved Error Handling

Buuhuu:
## Description

This PR handles runtime exceptions of the `GraphqlClient` by transforming them into valid error responses. 

All components have been changed to handle the error responses accordingly. 

## Related Issue

CIF-2078

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:


dplaton (approved):

dplaton (approved):

LSantha:
A side effect I see is that certain log messages may get duplicated with this PR. There is logging already in the graphql client implementation for graphql errors which were not logged in the components. Now that the runtime exception is converted to a graphql error while also logged and more logging is added in the components for graphql errors it may happen that new irrelevant warnings are logged in the components or an exception is logged as error and then a graphql error is logged as warning. 

Would it be possible to keep the distinction between real graphql errors coming from the graphql endpoint and other errors resulting from the access and communication problems related to the endpoint?

Buuhuu:
> A side effect I see is that certain log messages may get duplicated with this PR. There is logging already in the graphql client implementation for graphql errors which were not logged in the components.

There is also a duplicated logging in the graphql client, yes: https://github.com/adobe/commerce-cif-graphql-client/blob/master/src/main/java/com/adobe/cq/commerce/graphql/client/impl/GraphqlClientImpl.java#L239 ... it should be removed as the GraphqlClient is on an abstraction level where it should not care about the errors. If it would it should not log but throw a checked exception instead

> Would it be possible to keep the distinction between real graphql errors coming from the graphql endpoint and other errors resulting from the access and communication problems related to the endpoint?

I would not add this complexity for for logging purpose but rather reduce the logging levels either for the MagentoGraphqlClient or the Components. However the warning in the components are on the abstraction level of the component (for example failed to fetch category for identifier XYZ). The component doesn't care if it is caused by a runtime issue, a protocol error or whatever. The MagentoGraphqlClient though must log the exception (in the context of a query). So I would not argue that those are duplicated log messages.

Regarding log levels. Just because an graphql error is called error doesn't mean I have to log it as error. If a component still operates without intervention any exceptional case may be a warning. If it requires intervention it must be logged as error. For the RuntimeExceptions this is the case, thats why they are logged as error. That a component subsequently has to deal with an empty result set does not require any intervention and so it is logged as warning.

LSantha:
Hm, apparently this is a topic which would deserve a technical grooming or a more comprehensive analysis and design of what would be the correct long term approach. Adding check exceptions anywhere is a non backwards compatible API change and probably wouldn't be the solution anyway. The distinction between real graphql errors and other errors is not for logging only. Components may not need to emit any warning in certain situations. (If an author puts an invalid identifier in a component that probably shouldn't results in alerts about too many warnings in the logs etc.)

Buuhuu:
Right. We have another story pending for the "real" error reporting. As I don't have hard feeling pro/con logging the warnings or not I will just remove them. We can discuss a complete approach in the follow up topics.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/client/MagentoGraphqlClientImpl.java
```diff
@@ -270,6 +280,14 @@ private static GraphqlClient adaptToGraphqlClient(Resource resource) {
         return graphqlClient;
     }
 
+    private static GraphqlResponse<Query, Error> newErrorResponse(Throwable throwable) {
+        GraphqlResponse<Query, Error> response = new GraphqlResponse<>();
+        Error error = new Error();
+        error.setMessage(throwable.getMessage());
```
LSantha, line 286: One thing we may consider here is to use the category property of the error object to set a special predefined category when the error represents an exception instead of a GraphQL error which is its original purpose. That can be used then for special error handling scenarios if needed also by customers and can be part of the Javadoc of MagentoGrapqlClient.

Buuhuu: Thats a good idea. Will do that.

Buuhuu: Done in https://github.com/adobe/aem-core-cif-components/pull/641/commits/36299db78dcdbc1abd07485134d1f1c748477722

LSantha: Thanks you, LGTM.

LSantha (approved):

herzog31 (approved):

PR #642, opened 2021-07-29
CIF-2163: remove the usage of /magento/graphql

Buuhuu:
## Description

The default endpoint is `/api/graphql` for a while already. In this PR we remove the remaining occurrences of the former `/magento/graphql` endpoint. 

## Related Issue

CIF-2163

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/storeconfigexporter/StoreConfigExporterImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/storeconfigexporter/StoreConfigExporterImpl.java
@@ -20,6 +20,7 @@ import java.util.Map;
 
 import javax.annotation.PostConstruct;
 
+import org.apache.commons.collections4.MapUtils;
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.models.annotations.Model;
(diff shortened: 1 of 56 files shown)


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/storeconfigexporter/StoreConfigExporterImpl.java
```diff
@@ -53,31 +55,31 @@
...
     private Resource resource;
 
-    private String storeView;
-    private String graphqlEndpoint = "/magento/graphql";
+    private String storeView = DEFAULT_STORE_VIEW;
+    private String graphqlEndpoint = DEFAULT_GRAPHQL_ENDPOINT;
     private HttpMethod method = HttpMethod.POST;
     private Page storeRootPage;
-    private Map<String, String> httpHeaders;
+    private Map<String, String> httpHeaders = Collections.emptyMap();
```
herzog31, line 74: With that you can remove the null check in L108.

herzog31 (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/storeconfigexporter/StoreConfigExporterImpl.java
Buuhuu: good catch. replaced it with an `isEmpty()` to still return a static String instead of instantiating an `ObjectMapper` to generate an empty json object.

LSantha (approved):

PR #650, opened 2021-08-05
CIF-1502 - XF based page header with authoring support

LSantha:
Added version 2 for the following components to make them usable in the page editor: accountedtails, addressbook, minicart, resetpassword, searchbar, miniaccount.

## Related Issue

CIF-1502

## How Has This Been Tested?

Manually.

## Types of changes

## Checklist:


herzog31:
@LSantha What do you think about introducing a generic "Frontend Placeholder" component instead of having 5 individual placeholder components? The component could have a dialog with a select field to select between predefined placeholders (accountedtails, addressbook, minicart, resetpassword, miniaccount) and a free textfield to use a custom CSS class in case customers want to add additional frontend components.

LSantha:
I think from an authoring standpoint and also for future evolution the current approach is better. 
With a proper placeholder component the related react component has a proper representation and identity in the editor and in the authors mind. Instead of removing the placeholders, we could think for instance about what could be made configurable in each of those react components in the future and add options in the edit dialog of the component, but other usages could be considered too.

herzog31 (approved):

Buuhuu (approved):

PR #738, opened 2021-11-04
CIF-2437: implement url_path for products for all components

Buuhuu:
## Description

- This PR adds the url_path to all components it was missing so far
- It also implements a fallback to the url_rewrites in case the url_path (deprecated) is returned empty
- In order to make this work the `UrlProvider` api got extended to accept a type safe parameters object for product and category urls instead of the previous `Map<String,String>`, which would not have been able to provide the url_rewrites to the format implementations. The methods using `Map` got deprecated along the old `UrlFormat` interface. A new `GenericUrlFormat` replaces it with the two specific marker interfaces `ProductUrlFormat` and `CategoryUrlFormat`, both accepting only the parameters relevant for either (again compared to the `Map<String,String>` before)
- Compatibility of the old APIs and the new ones was implemented, only the fallback to the `url_rewrites` will not work in those cases

## Related Issue

CIF-2437

## Motivation and Context

Full support of url_path for products. 
First step for contextual aware product urls.

## How Has This Been Tested?

Unit Tests, locally using ProductPageUrlWithSkuAndUrlPath

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/services/UrlProviderImpl.java
```diff
@@ -115,121 +120,150 @@
...
-                ProductInterface product = retriever.fetchProduct();
-                if (product != null) {
-                    params
-                        .urlKey(product.getUrlKey())
-                        .urlPath(product.getUrlPath());
-                } else {
-                    LOGGER.debug("Could not generate product page URL for {}.", productIdentifier);
+            params.setSku(productIdentifier);
+            // assume that any other format then the ProductPageWithSku requires more parameters
+            if (!(newProductUrlFormat instanceof ProductPageWithSku)) {
```
Buuhuu, line 177: When we last refactored this we introduced the `UrlFormat#getParameterNames()` to allow `UrlFormats` to return which information they need to format a url. This allowed us to bypass the query here when only the sku is needed. However, this is also the only use case I could imagine. Any other format would require at least one more parameter. Due to that I removed the `getParameterNames()` again and replaced it with this `instanceof` check.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/services/sitemap/CategoriesSitemapGenerator.java
```diff
@@ -178,19 +174,25 @@ public void generate(Resource sitemapRoot, String name, Sitemap sitemap, Context
...
-                    .updatedAt()
-                    .children(child -> child.uid())));
+                .items(category -> {
+                    category
+                        .urlKey()
+                        .urlPath()
+                        .uid()
+                        .children(child -> child.uid());
+
+                    if (addLastModified) {
```
herzog31, line 191: Those fields are deprecated in the GraphQL API btw.

herzog31 (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/services/sitemap/CategoriesSitemapGenerator.java
Buuhuu: This change is unrelated to the ticket ... I missed that.

However we managed to convince the Magento team to keep those fields and un-deprecate :)

mhaack (approved):

PR #740, opened 2021-11-09
CIF-2544: remove usage of org.apache.sling.settings

Buuhuu:
## Description

This PR removes the usage of the SlingSettingsService. It moves this decision from runtime to build time by introducing a configureable component that is used by the `UrlProviderImpl` to decide if specific pages should be returned or not. As this logic is tightly coupled to the specific page handling of the `UrlProviderImpl` the related logic was moved the the new component as well. The `UrlProviderImpl` and `SpecificPageServlet` use this component now as service instead of static methods. 

## Related Issue

CIF-2544

## Motivation and Context

Remove API usage deprecated for cloud service usage.

## How Has This Been Tested?

Unit tests and locally

## Types of changes

## Checklist:


LSantha:
Would be possible to check in the URL provider that the commerce add-on bundle is available or not? It looks simpler than adding an extra OSGi config which is not supposed to be changed anyway and doing all the refactoring for the sake of this OSGi config.

Buuhuu:
Using runmode specific configuration is, as discussed the best practice. See also the adobetech article about it [here](https://medium.com/adobetech/custom-runmodes-on-aem-as-a-cloud-service-79b757f51a6b). 

I don't think adding an EnvironmentService is a good design however compared to making particular functionality configureable, because it will lead to n modules having n EnvironmentServices basically duplicating logic. Doing this for particular functionality (as here enable/disable deep linking to specific pages) causes duplication as well but it more expressive.

herzog31:
Related conversation: https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/2476

LSantha:
@Buuhuu , you could check that using BundleContext in UrlProviderImpl.activate().

Buuhuu:
@LSantha I know it is possible but it is not the right approach, see all the linked references.

herzog31 (approved):

PR #743, opened 2021-11-15
CIF-2455 - Add to cart of Product Teaser component fails for base products

LSantha:
* added call to action command in product teaser which may have various values for a call to action
 * open product details with add to cart button in product teaser when the product cannot be added directly to the cart
 * adjusted the product card in product recommendations to have the same behaviour as in product teaser
 * added unit tests

## Related Issue

CIF-2455

## How Has This Been Tested?

Manually, unit tests.

## Types of changes

- [B] Bug fix (non-breaking change which fixes an issue)
- [B] New feature (non-breaking change which adds functionality)

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productteaser/ProductTeaserImpl.java
```diff
@@ -173,6 +174,22 @@ public String getCallToActionText() {
         return ctaText;
     }
 
+    @Override
+    public String getCallToActionCommand() {
```
herzog31, line 178: Instead of adding a new public method, I would do the check in the `getCallToAction` method and return a new value like `add-to-cart-redirect` which can then indicate for HTL to render an "add to cart" button which however links to the product detail page.

ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/productteaser/v1/productteaser/actions.html
```diff
@@ -15,11 +15,11 @@
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
 <template data-sly-template.actions ="${@ product}">
     <div class="productteaser__cta" data-sly-test="${product.callToAction != null}">
-        <button data-sly-test="${product.callToAction == 'add-to-cart'}" data-action="addToCart" data-item-sku="${product.sku}"
-                class="button__root_highPriority button__root clickable__root button__filled" type="button">
+        <button data-sly-test="${product.callToAction == product.CALL_TO_ACTION_TYPE_ADD_TO_CART}" data-action="${product.callToActionCommand}" data-item-sku="${product.sku}"
```
Buuhuu, line 18: This will not resolve the defect for customers that have overwritten the `actions.html`. We should try to find a solution that works for all customers.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/productteaser/ProductTeaser.java
```diff
@@ -84,6 +88,21 @@
      */
     String getCallToActionText();
 
+    /**
+     * Returns the call to action command for the configured call to action.
+     *
+     * @return "addToCart" or "details" or {@code null} if call to action is not configured
+     */
+    default String getCallToActionCommand() {
```
Buuhuu, line 96: A new public method is unnecessary to resolve this neither is a change to the htl necessary. You can achieve the same outcome without touching the public interface. 

See the following implementation only proposal
```@Override
public String getCallToAction() {
    if (StringUtils.isNotEmpty(cta)) {
        // return always details if the product cannot be added to cart (e.g. configurable product)
        return oneClickShoppable(getProduct()) ? cta : "details";
    }
    return null;
}

@Override
public String getCallToActionText() {
    if (StringUtils.isNotEmpty(ctaText)) {
        return ctaText;
    } else if ("details".equals(cta)) {
        return getI18n().get("See more details");
    } else if ("add-to-cart".equals(cta)) {
        return getI18n().get("Add to Cart");
    }
    return null;
}
```

Buuhuu (changes requested):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productteaser/ProductTeaserImpl.java
LSantha: I thought of this but it would add more duplication to the markup, so I looked for other options.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/productteaser/ProductTeaser.java
LSantha: This seems to fix the issue with minimal impact. The logic behind it is quite tricky that one may not want to multiply in the future but may work for the current case.

Buuhuu: If we have other cases in the future, it makes sense to revisit it but now we dont. As discussed, doing a minimal impact change for a bug that will effectively address the issue for all customers (no matter of `actions.html` is overwritten or not) seems reasonable reasonable and more favourable.   

If it is _tricky_ or not is a subjective opinion. The "command" being not-equal to the "action" is tricky to understand either + it requires more logic on the client to figure out what to finally do.

herzog31 (approved):

Buuhuu (approved):

PR #764, opened 2021-11-25
CIF-2574 - Transformer support for product / category title as commerce link text

LSantha:
* replace link contents with product or category title if attribute data-replace-text is true
 * set title to product or category title if attribute data-replace-text is true and title is not set
 * updated unit tests

## Related Issue

CIF-2574

## How Has This Been Tested?

JUnit, manually.

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/services/CommerceLinksTransformerFactory.java
```diff
@@ -115,25 +139,147 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+                    if (title == null) {
+                        // set title to linkText
+                        newAttributes.addAttribute("", ATTR_TITLE, ATTR_TITLE, "CDATA", linkText);
+                    }
+                }
+                super.startElement(uri, localName, qName, newAttributes);
+                if (StringUtils.isNotBlank(linkText)) {
+                    char[] chars = linkText.toCharArray();
+                    // insert linkText as new content
+                    super.characters(chars, 0, chars.length);
```
Buuhuu, line 181: You could also write the end endElement here directly. So you would not have to keep track on the nesting level of the <a> tags descendants but could simply ignore all up to and including the original endElement tag.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/services/CommerceLinksTransformerFactory.java
```diff
@@ -115,25 +139,147 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+            if (product == null) {
+                LOGGER.debug("Product not found for SKU {}.", productSku);
+                return null;
             }
 
-            super.startElement(uri, localName, qName, attributesImpl);
+            // set link text to product name
+            linkText = product.getName();
+
+            ProductUrlFormat.Params params = new ProductUrlFormat.Params();
```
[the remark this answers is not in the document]
LSantha: Good point. I've fixed it.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/services/CommerceLinksTransformerFactory.java
LSantha: I don't see how that would work. The endElement() will be called anyway when the parsing reaches </a> of the commerce link and then we need to stop ignoring content by tracking the nesting level.

Buuhuu (approved):

mhaack (approved):

PR #766, opened 2021-11-29
CIF-2566: Add missing datalayer events

laurentiumagureanu:
## Description

Added missing cmp:clicked component events

## Related Issue

CIF-2566

## Motivation and Context

Generate adobeDataLayer events

## How Has This Been Tested?

manually

## Types of changes

## Checklist:


react-components/src/utils/useDataLayerEvents.js
```diff
@@ -0,0 +1,51 @@
...
+
+            if (item.bundle) {
+                eventInfo['bundle'] = true;
+            }
+
+            dataLayerUtils.pushEvent('cif:addToCart', eventInfo);
+        });
+    });
+
+    useEventListener(document, 'aem.cif.add-to-wishlist', async event => {
```
Buuhuu, line 37: This change should be in a separate PR based on issue/CIF-2407.

Buuhuu (commented):
What about product recommendation's open details? Do we need the click event there as well?

react-components/src/utils/useDataLayerEvents.js
laurentiumagureanu: not necessarily. it has no side-effects if the event is never triggered

Buuhuu: In the base of this PR the even `aem.cif.add-to-wishlist` does not exist. for traceability it would be better to have this done as part of CIF-2407. However, if we merge both in the same release it is probably not worth the effort to move the change.

react-components/src/utils/useDataLayerEvents.js
```diff
@@ -0,0 +1,51 @@
...
+                'xdm:SKU': item.sku,
+                'xdm:quantity': item.quantity
+            };
+
+            dataLayerUtils.pushEvent('cif:addToWishList', eventInfo);
+        });
+    });
+};
+
+export default useDataLayerEvents;
```
herzog31, line 51: What do you think about exporting this in the index as well, so it can be used directly without the need to use the `App` component?

herzog31 (approved):

react-components/src/utils/useDataLayerEvents.js
laurentiumagureanu: done 👍

Buuhuu (approved):

PR #769, opened 2021-12-01
CIF-2577: add wish list to gift card products

Buuhuu:
## Description

In this PR the gift card product details got extended by an add to wish list button.

## Related Issue

CIF-2407

## How Has This Been Tested?

Locally and existing unit tests

## Types of changes

## Checklist:


react-components/src/components/GiftCardOptions/giftCardOptions.js
```diff
@@ -48,13 +49,21 @@ const messages = defineMessages({
...
+            changeOptionValue,
+            changeQuantity,
+            canAddToCart,
+            addToCart,
+            addToWishlist
+        }
     ] = useGiftCardOptions({ sku });
 
-    const { formatMessage } = useIntl();
+    const intl = useIntl();
```
herzog31, line 66: Is this change necessary?

herzog31 (approved):

react-components/src/components/GiftCardOptions/giftCardOptions.js
Buuhuu: It is actually recommended to use `intl` that way to ease extraction. See the https://formatjs.io/docs/tooling/cli example

```
function Comp(props) {
  const {intl} = props
  return intl.formatMessage({
    // The whole `intl.formatMessage` is required so we can extract
    id: 'foo',
    defaultMessage: 'foo',
    description: 'bar',
  })
}
```

react-components/src/components/GiftCardOptions/useGiftCardOptions.js
```diff
@@ -25,43 +25,41 @@ const useGiftCardOptions = props => {
...
-            if (!sku) {
-                console.error('SKU is not present in the dataset of mountpoint element');
-                return;
-            }
-            const { data, error } = await giftCardProductQuery({ variables: { sku }, fetchPolicy: 'network-only' });
-
-            if (error) {
-                throw new Error(error);
-            }
+    const fetchGiftCardOptions = useCallback(async () => {
```
laurentiumagureanu, line 28: Why is this change necessary ?

Buuhuu: This returns a function that is updated when the `sku` changes. And again when the function changes `useEffect` is called again to fetch the gift card options. Before the options were fetched only once. 

I used `useCallback` to create a memorised value as it is unlikely that the `sku` actually changes.

laurentiumagureanu: why not use just the `useEffect` with a dependency on `sku` ?

Buuhuu: done

laurentiumagureanu (approved):

PR #781, opened 2022-01-06
CIF-2618: make StoreConfigExporter a ConsumerType

Buuhuu:
## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java
@@ -81,6 +81,15 @@ public interface MagentoGraphqlClient {
      * Returns the list of custom HTTP headers used by the GraphQL client.
      *
      * @return a {@link Map} with header names as keys and header values as values
+     * @deprecated this method will be replaced by {@link MagentoGraphqlClient#getHttpHeaderMap()} in the next major release‡
      */
+    @Deprecated
     Map<String, String> getHttpHeaders();
+
+    /**
+     * Returns the Map of custom HTTP headers used by the GraphQL client.
+     * 
+     * @return a {@link Map} with header names as keys and header values as values
+     */
+    Map<String, String[]> getHttpHeaderMap();
 }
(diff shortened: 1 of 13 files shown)


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java
```diff
@@ -81,6 +81,15 @@
...
+    @Deprecated
     Map<String, String> getHttpHeaders();
+
+    /**
+     * Returns the Map of custom HTTP headers used by the GraphQL client.
+     * 
+     * @return a {@link Map} with header names as keys and header values as values
+     */
+    Map<String, String[]> getHttpHeaderMap();
 }
```
mhaack, line 95: While I don't disagree with this improvement, how is that related to the PR goal of making StoreConfigExporter a ConsumerType?

mhaack (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/client/MagentoGraphqlClient.java
Buuhuu: It is related to the change in the StoreConfigExporter interface. 

Initially we used the `StoreConfigExporterImpl` in the `PageImpl` because we wanted to have a method that exposes the http headers as Map instead of an already serialised object. This did not require a change to the interface at this time.

However, now that we use the interface we had to add this method to the interface as well and as headers can appear more than once using a `Map<String, String[]>` is the right return value for that. (See also the servlet spec). As the Map of headers returned by the StoreConfigExporter is directly derived from the MagentoGraphqlClient, I created the method that returns the data in the same, correct format.

Bottom line: this has been a design mistake in the first place and this is an attempt to clean things up in as less breaking iterations as possible.

mhaack: > Bottom line: this has been a design mistake in the first place and this is an attempt to clean things up in as less breaking iterations as possible.

That was my thinking as well for question. :-)

PR #797, opened 2022-01-24
CIF-2644 - Update MSE and recommendations to support more page types

herzog31:
## Description

* Extended `usePageType` hook to support additional page types. The page types are used for Magento storefront events and product recommendations.
* The following page types are supported now:
    * Homepage
    * Content page
    * Product page
    * Category page
    * Cart page
    * Checkout page

## How Has This Been Tested?

* Updated unit tests

## Types of changes

## Checklist:

### DIFF
--- a/react-components/src/utils/hooks.js
+++ b/react-components/src/utils/hooks.js
@@ -134,7 +134,7 @@ export const usePageType = () => {
         // Could not parse store config, ignore for now
     }
 
-    if (template == 'landing-page' || (canonicalUrl && storeRootUrl && canonicalUrl == storeRootUrl)) {
+    if (template == 'landing-page' || (canonicalUrl && storeRootUrl && canonicalUrl.endsWith(storeRootUrl))) {
         return PageTypes.CMS;
     }
     if (document.querySelector('[data-cif-product-context]') || template == 'product-page') {
--- a/react-components/src/utils/hooks.js
+++ b/react-components/src/utils/hooks.js
@@ -94,7 +94,6 @@ export const useStorefrontEvents = () => {
 };
 
 export const usePageType = () => {
-
     const PageTypes = {
         // Pages with landing-page template or pages that match the store root URL
         // as defined in the store config.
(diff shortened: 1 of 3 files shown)


react-components/src/utils/hooks.js
```diff
@@ -94,21 +94,61 @@ export const useStorefrontEvents = () => {
...
+
+    let storeRootUrl = null;
+    try {
+        let storeConfig = JSON.parse(document.querySelector('meta[name="store-config"]').getAttribute('content'));
+        storeRootUrl = storeConfig.storeRootUrl;
+    } catch (err) {
+        // Could not parse store config, ignore for now
+    }
+
+    if (template == 'landing-page' || (canonicalUrl && storeRootUrl && canonicalUrl == storeRootUrl)) {
```
Buuhuu, line 137: This assumption `canonicalUrl == storeRootUrl` is not correct. 

The storeRootUrl is [mapped per default](https://github.com/adobe/aem-core-cif-components/blob/master/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/storeconfigexporter/StoreConfigExporterImpl.java#L118) using `ResourceResolve#map(HttpServletRequest, String)` and canonical url using `ResourceResolver#map(String)` (without the request context). 

The storeRootUrl may be a path `/us/en.html` but the canonical url will always be absolute `http://wknd.site/en.html`.

herzog31: This seems to be inconsistent. If you check https://publish-p10603-e75517-cmstg.adobeaemcloud.com/us/en.html, the canonical URL is relativ. When going to any other page, it becomes absolute.

Any idea how else we can check which page is the homepage? Template name alone might not be sufficient, since this is custom and the landing-page template is only a recommendation.

Buuhuu: This inconsistency is addressed in https://github.com/adobe/aem-cif-guides-venia/pull/245

I guess you could just remove the authority of the canonical url and only compare the paths.

Buuhuu (approved):

PR #821, opened 2022-02-11
CIF-2519 - Support category filter for product list component

LSantha:
* added support for manual category selection in edit dialog and model
 * added unit tests
 * added sample page in examples project for manual category selection
 * added integration test

## Related Issue

CIF-2519 

## How Has This Been Tested?

Manual, unit test, integration test, UI test.

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImpl.java
```diff
@@ -75,6 +76,7 @@
 
     private static final boolean SHOW_TITLE_DEFAULT = true;
     private static final boolean SHOW_IMAGE_DEFAULT = true;
+    private static final String SELECTION_PROPERTY = "selection";
```
Buuhuu, line 79: Just a minor thing about the naming: I would prefer if we could call that `category`. If we want to add additional properties in the future that impact the "selection" having the category being called `selection` may become ambiguous. 

We had a similar thing with the way the specific page properties are configured and had to ultimately duplicated the category picker field because of it.

Buuhuu (approved):

mhaack (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/productlist/ProductListImpl.java
LSantha: Hm, I was thinking about that and eventually went for the same name as the Product component with some doubts. I can change it if needed.

Buuhuu: If we could, we should change the product component as well, but we cannot due to backward compatibility reasons. (same as for the specific page properties case mentioned before).

I think we should really be more specific with the property names to guarantee a forward compatibility. However, I understand your reason to keep it aligned with what was there in other components.

LSantha: OK, I've changed the name.

PR #884, opened 2022-04-12
CIF-1696: Migrate add to cart code to new generic addProductsToCart mutation

laurentiumagureanu:
## Description

Added a Product v3 component that includes the UIDs of configurable options required by the new mutation
Adapted the JS code to use those attributes for variant selection and add to cart

## Related Issue

CIF-1696

## Motivation and Context

Enables usage of new generic addProductsToCart mutation

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/product/Variant.java
```diff
@@ -44,5 +45,9 @@ public interface Variant {
 
     Map<String, Integer> getVariantAttributes();
 
+    default Map<String, String> getVariantAttributesUid() {
```
Buuhuu, line 48: The implementation expects the returned map to be mutable
```
mappedVariant.getVariantAttributesUid().put(option.getCode(), option.getUid().toString());
```
Returning an empty map here will throw an exception with the above statement if not correctly implemented by customers.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v3/product/ProductImpl.java
```diff
@@ -0,0 +1,91 @@
...
+        attribute.setLabel(option.getLabel());
+        attribute.setId(option.getAttributeCode());
+        attribute.setValues(values);
+
+        return attribute;
+    }
+
+    @Override
+    protected Variant mapVariant(ConfigurableVariant variant) {
+        Variant mappedVariant = super.mapVariant(variant);
```
Buuhuu, line 82: Maybe it would be better to return the `VariantImpl` from the `mapVariant()` call and add a method to it to mutate the internal state of the `variantAttributesUid`.  Ideally the collections/maps returned by a data object are immutable.

react-components/src/talons/Cart/useAddToCartEvent.js
```diff
@@ -41,10 +41,23 @@ const useAddToCartEvent = (props = {}) => {
 
     useEventListener(document, 'aem.cif.add-to-cart', async event => {
         const items = typeof event.detail === 'string' ? JSON.parse(event.detail) : event.detail;
-        const physicalCartItems = items.filter(item => !item.virtual).map(productMapper);
-        const virtualCartItems = items.filter(item => item.virtual).map(productMapper);
-        const bundleCartItems = items.filter(item => item.bundle).map(bundledProductMapper);
-        const giftCardCartItems = items.filter(item => item.giftCard).map(giftCardProductMapper);
+        const useUidItems = items
```
Buuhuu, line 44: trivial but it would be good to have a comment here explaining what the difference between the items with `useUid` is and why we keep the "old way of things".

Buuhuu (approved):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v3/product/ProductImpl.java
laurentiumagureanu: Agreed. But I was following the same pattern applied for `getVariantAttributes`.

ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/product/v1/product/clientlib/js/variantSelector.js
```diff
@@ -45,6 +45,10 @@ class VariantSelector {
             button.addEventListener('click', this._onSelectVariant.bind(this));
         }, this);
 
+        this._state.useUid = this._element
```
Buuhuu, line 48: I get an error here, which is caused by `this._element.querySelector(VariantSelector.selectors.quantity)` returning `null`

```
clientlib-cif.lc-8a1ac571e177293e65b2067774853a93-lc.min.js:1030 Uncaught TypeError: Cannot read properties of null (reading 'hasAttribute')
    at new VariantSelector (clientlib-cif.lc-8a1ac571e177293e65b2067774853a93-lc.min.js:1030:89)
    at HTMLDocument.onDocumentReady (clientlib-cif.lc-8a1ac571e177293e65b2067774853a93-lc.min.js:1209:29)
```

react-components/src/components/BundleProductOptions/bundleProductOptions.js
```diff
@@ -36,7 +36,7 @@ const BundleProductOptions = () => {
     const productId = document.querySelector('[data-cmp-is=product]')?.id;
     const [bundleState, setBundleState] = useState(null);
     const intl = useIntl();
-    let { sku, showAddToWishList } = document.querySelector(bundleProductOptionsContainer)?.dataset || {};
+    let { sku, showAddToWishList, useUid } = document.querySelector(bundleProductOptionsContainer)?.dataset || {};
```
Buuhuu, line 39: useUid will be initialised with `''` which later evaluates to false as an empty string is a falsy value. This should be checked for undefined to cast it to a proper boolean

```
useUid = typeof useUid !== 'undefined'
```

Issue #886, opened 2022-04-27
URL Provider Update - Allow for querying category by URL Path


pmichelotti:
---
name: Allow for querying category by URL Path
about: URL Provider
---

### User Story
As a site manager I want to be able to set category ID lookup to be based off URL Path so that categories with matching URL keys can be looked up uniquely

### Description & Motivation
In large enough catalog taxonomies it is fairly common for URL keys to match between categories.  The current URL Provider implementation uses URL Key to look up the ID of a category which causes effectively indeterminate behavior when there are categories with matching URL Keys (see https://github.com/adobe/aem-core-cif-components/blob/master/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/services/UrlProviderImpl.java#L451).  

### Deliverables
Code change needed to update the URL Provider to know whether to use URL Key or URL Path to query for category IDs.

### Acceptance Criteria
* I am able to set in the configuration of the URL Provider whether URL Path or URL Key or ID should be used to identify categories
* Given that I am viewing two categories with matching URL Keys but different URL Paths and that I have configured the URL Provider to lookup categories by URL Path and to render URL Paths in the URL then I see the expected category content in each category I am viewing

### Verification Steps
* Setup multiple categories in Magento with the same URL key at different places in a taxonomy so the URL path is differing
* Configure the URL provider to user URL path to lookup category ID
* Open category pages for both categories and make sure the content matches what is expected based on the category definitions in Magento

Buuhuu:
@adobe export issue to Jira project CIF

Buuhuu:
Released with CIF Core Components 2.9.0

PR #899, opened 2022-05-03
CIF-2864: support querying category by url_path

Buuhuu:
## Description

In order to support setups where multiple categories share the same `url_key` we have to resolve the category uid from the url path when the format is configured accordingly. 

The change requires additional query support by 3rd party integrations and may even be breaking if they depend on the query by `url_key` but use the category with url path url format.

## Related Issue

CIF-2864
#[886](https://github.com/adobe/aem-core-cif-components/issues/886)

## How Has This Been Tested?

Locally and with unit tests

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/services/UrlProviderImpl.java
```diff
@@ -442,20 +442,27 @@ public String getCategoryIdentifier(SlingHttpServletRequest request) {
 
         if (StringUtils.isNotEmpty(categoryIdentifiers.getUid())) {
             identifier = categoryIdentifiers.getUid();
-        } else if (StringUtils.isNotEmpty(categoryIdentifiers.getUrlKey())) {
-            // lookup internal product identifier (sku) based on URL product identifier
-            // (url_key)
+        } else {
+            // lookup internal category identifier (uid) based on URL category identifier (url_key)
```
herzog31, line 446: The comment is now outdated. It's `url_key` or `url_path`.

herzog31 (approved):

PR #912, opened 2022-05-24
CIF-2836: product color options as css classes

laurentiumagureanu:
## Description

Uses the label of the configurable product options to add class modifiers on option tiles in the product component

## Related Issue

CIF-2836

## Motivation and Context

Adds the ability to better style the product options tiles (colors, sizes, etc).

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/VariantValueImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/VariantValueImpl.java
@@ -16,7 +16,6 @@
 package com.adobe.cq.commerce.core.components.internal.models.v1.product;
 
 import com.adobe.cq.commerce.core.components.models.product.VariantValue;
-import com.adobe.cq.commerce.magento.graphql.SwatchDataInterface;
 
 public class VariantValueImpl implements VariantValue {
 
@@ -28,7 +27,7 @@ public class VariantValueImpl implements VariantValue {
 
     private String uid;
 
-    private SwatchDataInterface swatchData;
+    private SwatchType swatchType;
 
     @Override
     public String getLabel() {
@@ -40,7 +39,7 @@ public class VariantValueImpl implements VariantValue {
     }
 
     @Override
-    public String getDefaultLabel() {
+    public String getCssClassModifier() {
         return defaultLabel;
     }
 
@@ -49,12 +48,12 @@ public class VariantValueImpl implements VariantValue {
     }
 
     @Override
-    public SwatchDataInterface getSwatchData() {
-        return swatchData;
+    public SwatchType getSwatchType() {
+        return swatchType;
     }
 
-    public void setSwatchData(SwatchDataInterface swatchData) {
-        this.swatchData = swatchData;
+    public void setSwatchType(SwatchType swatchType) {
+        this.swatchType = swatchType;
     }
 
     @Override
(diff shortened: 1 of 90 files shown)


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/product/VariantValue.java
```diff
@@ -25,9 +27,17 @@
...
+        return null;
+    }
+
     Integer getId();
 
     default String getUid() {
         return null;
     }
+
+    default SwatchDataInterface getSwatchData() {
```
Buuhuu, line 40: We should not add a dependency to the SwatchDataInterface (DTO) here. This would couple the API to closely to the gql binding and if we have to make changes we cannot as we don't own the interface.

Ideally the `Variant` interface defines it's type, for example as an enum: `COLOR`, `TEXT`. This could be set in the model and be read in the htl script accordingly without exposing the DTO.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/product/VariantValue.java
```diff
@@ -25,9 +27,17 @@
 public interface VariantValue {
     String getLabel();
 
+    default String getDefaultLabel() {
```
Buuhuu, line 30: Please use a more descriptive name that matches the purpose of the value: `getCssClassName()`

Buuhuu (changes requested):

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/product/VariantValue.java
laurentiumagureanu: In this case we use it for class modifiers. but that is not the purpose of this field. we might end-up using this for something different in the future. I would keep it as is

Buuhuu: If we would use it as is (without any logic applied) I agree. In this particular case however we change the default label value in order to make it usable as BEM modifier in a css class name. If we leave it with `getDefaultLabel()` and we would want to expose the real default label in the future, how would you name the method? 

Let's find a better naming that reflects what the values is meant for.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/product/VariantValueImpl.java
```diff
@@ -34,6 +38,24 @@ public void setLabel(String label) {
         this.label = label;
     }
 
+    @Override
+    public String getCssClassModifier() {
+        return defaultLabel;
+    }
+
+    public void setDefaultLabel(String defaultLabel) {
```
Buuhuu, line 46: java bean getter and setters should be named by the field `cssClassModifer`, `getCssClassModifier()`, `setCssClassModifier()`

ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/product/v1/product/variantselector.html
```diff
@@ -15,16 +15,18 @@
...
             <div
-                data-sly-test.isColor="${variantAttribute.label == 'Color'}"
+                data-sly-test.isColor="${isColorOption || variantAttribute.label == 'Color'}"
                 data-id="${variantAttribute.id}"
                 class="swatchList__root tileList__root"
             >
                 <sly data-sly-list.variantValue="${variantAttribute.values}">
                     <button
-                        class="swatch__root tile__root clickable__root"
+                        class="${'swatch__root clickable__root tile__root{0}{1}' @ format=[hasSwatchData ? ' tile__root--' : '', hasSwatchData ? variantValue.cssClassModifier : ''], context='text'}"
```
Buuhuu, line 29: The modifier should be applied also for attributes that don't have swatch data but have the label 'Color'.

laurentiumagureanu: if there is no swatch data then there is no css class modifier in the variant value

LSantha (approved):

herzog31 (approved):

PR #926, opened 2022-06-21
CIF-2738: support client side price loading in product teaser and carousel

Buuhuu:
## Description

To support clientside price loading for product teasers and product carousels a new `data-product-base-sku` attribute was added to the product teaser and each item of the product carousel. This was necessary as both components can be configured with variants but querying the prices of those requires the base product' sku. To provide this attribute the `ProductTeaser` interface and `ProductListItem` interface (used by the product carousel) have a new getCombinedSku() method that returns the sku as object containing the base sku and variant sku. 

Both components were also extended to query the maximum price range for bundled products to have a consistent result of serverside and client side price rendering.

Prices are loaded for all teasers and all carousels on a single page at once.

This change also changes the approach to enable client-side price loading site-wide using the CIF Cloud Configuration. There were no changes made to the policy properties of the product teaser and product carousel. The "Load prices" property of product v3 was removed. The per-component policy configuration for product v1 and v2 and the for the product list and search results remains working as before but the respective java API got deprecated.

Additional changes:
- move the conversion from GQL response to price model to CommerceGraphqlApi (removes duplication)
- move the price model to price html formatting to PriceFormatter (removes duplication)
- suspend `getProductPrices()` calls until all `aem.cif.clientlib-initialized` completed. Calls to `getProductPrices()` on document ready of the respective components (synchronous event listener) are queued, merged and executed at once. This reduces the graphql calls necessary on a page with many teasers, carousels or multiple product detail components with different visible sections to a maximum of two (one with variants another one without to prevent over fetching). Is cases where clientlibs are loaded async the `aem.cif.clientlib-initialized` may be dispatched before the component clientlib is loaded. In this case the `getProductPrices()` query individual as usual.

## Related Issue

https://github.com/adobe/aem-cif-guides-venia/pull/296

## How Has This Been Tested?

Unit tests and locally

## Types of changes

## Checklist:


herzog31 (approved):

ui.apps/src/main/content/jcr_root/apps/core/cif/clientlibs/common/js/PriceFormatter.js
```diff
@@ -17,11 +17,80 @@
 
 class PriceFormatter {
     constructor(locale) {
-        this._locale = locale;
+        this._locale = locale || window.CIF.locale || navigator.language;
```
herzog31, line 20: You can also use `document.documentElement.lang` as a fallback. That's the setting coming from the AEM page properties.

herzog31 (approved):

PR #931, opened 2022-06-24
CIF-2921: improve authoring experience of product v3 component

buuhuu:
## Description

This change replaces the many checkboxes with a multi select and enables it in the component dialog as well as in the design dialog to ease the use of the feature. 

Furthermore each section was extended with a per default hidden area title using the same markup as we have it already in place for some of the existing areas. This is in particular useful when a single section is show within the context of another component like a content fragment, where each element of the CF has a title and value. 

Additionally: 
- The cartActions css class was renamed to actions for the v3 product, but not in the bundleProductOptions.js and the giftCardOptions.js. This change adds the class also there additionally to the old one for backward compatibility reasons to the product v1 and v2
- I also added th `productFullDetail__options` to the bundle and gift card mountpoints in to allow them to use the same grid css styling as the other "options"-sections (simple/configurable & grouped product)

## Related Issue

CIF-2921

## Motivation and Context

Simplify the configuration of composable commerce experience 

## How Has This Been Tested?

Unit tests and locally with wknd

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/breadcrumb/BreadcrumbImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/breadcrumb/BreadcrumbImpl.java
@@ -40,16 +40,21 @@ import org.apache.sling.models.annotations.via.ForcedResourceType;
 
 import com.adobe.cq.commerce.core.components.client.MagentoGraphqlClient;
 import com.adobe.cq.commerce.core.components.internal.datalayer.DataLayerComponent;
+import com.adobe.cq.commerce.core.components.internal.services.site.SiteStructureImpl;
 import com.adobe.cq.commerce.core.components.internal.services.urlformats.UrlFormatBase;
 import com.adobe.cq.commerce.core.components.models.breadcrumb.Breadcrumb;
+import com.adobe.cq.commerce.core.components.models.common.SiteStructure;
 import com.adobe.cq.commerce.core.components.models.navigation.Navigation;
 import com.adobe.cq.commerce.core.components.services.urls.CategoryUrlFormat;
 import com.adobe.cq.commerce.core.components.services.urls.ProductUrlFormat;
+import com.adobe.cq.commerce.core.components.services.urls.UrlFormat;
 import com.adobe.cq.commerce.core.components.services.urls.UrlProvider;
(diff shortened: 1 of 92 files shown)


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v3/product/ProductImpl.java
```diff
@@ -37,36 +41,54 @@
...
+
         if (productRetriever != null) {
             productRetriever.extendProductQueryWith(p -> p.onConfigurableProduct(cp -> cp
                 .configurableOptions(o -> o
                     .values(v -> v.uid()))
                 .variants(v -> v
                     .attributes(a -> a.uid()))));
         }
+
+        if (visibleSections == null || visibleSections.length == 0) {
```
herzog31, line 88: When using the edit and design dialog together to define a setting, there are usually two aspects you want to cover:

- In the design dialog:
    - Define which options are available for the author in the edit dialog.
    - Define the default setting.
- In the edit dialog:
    - Customize component if you want to deviate from the default and choose between allowed settings.

This pattern is used in the WCM title component: https://github.com/adobe/aem-core-wcm-components/tree/main/content/src/content/jcr_root/apps/core/wcm/components/title/v3/title. I recommend verifying this with our PM.

ui.apps/src/main/content/jcr_root/apps/core/cif/components/commerce/product/v3/product/_cq_dialog/.content.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jcr:root xmlns:sling="http://sling.apache.org/jcr/sling/1.0" xmlns:cq="http://www.day.com/jcr/cq/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0" xmlns:nt="http://www.jcp.org/jcr/nt/1.0"
+    jcr:primaryType="nt:unstructured"
+    jcr:title="Product Detail"
+    sling:resourceType="cq/gui/components/authoring/dialog"
+    extraClientlibs="[cif.shell.picker]"
+    trackingFeature="cif-core-components:product:v1">
```
LSantha, line 7: should be v3

LSantha (approved):

PR #938, opened 2022-07-07
CIF-2895 - Extend product retriever with hook for product filters

herzog31:
## Description

* Added `extendCategoryFilterWith` and `extendProductFilterWith` to retrievers that support hooks to enable full replacement or extension of category and product query filters.
* Added unit tests.

Requires a release of https://github.com/adobe/commerce-cif-magento-graphql/pull/29 before it can be merged.

## Related Issue

See https://github.com/adobe/commerce-cif-magento-graphql/pull/27

## How Has This Been Tested?

* Unit tests + Venia example

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/retriever/AbstractProductRetriever.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/retriever/AbstractProductRetriever.java
@@ -17,6 +17,7 @@ package com.adobe.cq.commerce.core.components.models.retriever;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 
 import org.apache.commons.collections4.CollectionUtils;
@@ -35,6 +36,7 @@ import com.adobe.cq.commerce.magento.graphql.Query;
 import com.adobe.cq.commerce.magento.graphql.QueryQuery;
 import com.adobe.cq.commerce.magento.graphql.SimpleProductQuery;
 import com.adobe.cq.commerce.magento.graphql.gson.Error;
+import com.shopify.graphql.support.Input;
 
 /**
  * Abstract implementation of product retriever that loads product data using GraphQL.
@@ -50,6 +52,11 @@ public abstract class AbstractProductRetriever extends AbstractRetriever {
      * Lambda that extends the product variant query.
      */
     protected Consumer<SimpleProductQuery> variantQueryHook;
+
+    /**
+     * Lambda that replaces the product attribute filters.
+     */
+    protected BiConsumer<String, ProductAttributeFilterInput> productAttributeFilterHook;
 
     /**
      * Product instance. Is only available after populate() was called.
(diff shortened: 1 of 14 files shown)


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/retriever/AbstractProductRetriever.java
```diff
@@ -138,15 +145,47 @@ public void extendVariantQueryWith(Consumer<SimpleProductQuery> variantQueryHook
...
+     *     .setSku(new FilterEqualTypeInput()
+     *         .setEq(identifier))
+     *     .setCustomFilter("eco-friendly", new FilterEqualTypeInput()
+     *         .setEq("yes")));
+     * }
+     * </pre>
+     *
+     * @param productAttributeFilterHook Lambda that defined the product attribute filter.
+     */
+    public void replaceProductFilterWith(BiConsumer<String, ProductAttributeFilterInput> productAttributeFilterHook) {
```
Buuhuu, line 165: Could we do an extend...With approach here as well, similar to the query definition? 

Iiuc it is either the default or the custom filter. With this approach a custom implementation would need to duplicate the logic that would otherwise be set by default filter. Ideally they could just extend the default filter, e.g.

```
productRetriever.extendProductFilterWith((identifier, f) -> f
  .setCustomFilter("eco-friendly", new FilterEqualTypeInput()
    .setEq("yes")));
```

In this case the identifier argument would not be needed as any data to extend the filter would usually be owned by the extension anyway. 

To allow still implementations to replace the entire filter with a custom one we could use a `Function<ProductAttributeFilterInput,ProductAttributeFilterInput>` instead.

herzog31: IMO it's more likely to completely replace a filter than to extend it. For example for third-party implementations where you don't want to filter by a SKU, but by something else. Wrt CIF-2916, this can be handled as implementation detail of the retrievers. If customers wanted to customize this, they can get the hook from the UrlProvider and chain it together with their custom filter. In this case, we don't need to pass the `identifier`. 

So the question remains, what are we doing until then? One approach could be to only allow extending filters for now and once we have CIF-2916 we additionally allow to replace them, but without passing the identifier.

```java
// Today
productRetriever.extendProductFilterWith(f -> f
    .setCustomFilter("eco-friendly", new FilterEqualTypeInput()
        .setEq("yes")));


// With CIF-2916
Consumer<ProductAttributeFilterInput> filterHook = urlProvider.getFilterHook(request);
Consumer<ProductAttributeFilterInput> customHook = f -> f
    .setCustomFilter("eco-friendly", new FilterEqualTypeInput()
        .setEq("yes"));

productRetriever.replaceProductFilterWith(filterHook.andThen(customHook));
```

Buuhuu: > IMO it's more likely to completely replace a filter than to extend it. For example for third-party implementations where you don't want to filter by a SKU, but by something else.

IMO it is unlikely that a custom component has to replace the filter. The use case of the customer requesting this is not to "not filter by sky" but to filter by sku and additionally some custom criteria. We will most likely always filter by an identifier, as this is fundamental to how the retrievers work. As said, if I extend an existing component I don't want to duplicate the core component logic. I don't care if a product retriever filters by sku or url key and where it gets this information from. I just want to extend the constraint with my own logic. 

> If customers wanted to customize this, they can get the hook from the UrlProvider and chain it together with their custom filter. 
This adds unnecessary complexity for custom implementations. Also it would hurt adaption as we cannot expect customers to implement such a change in their custom component right away. If we do them in the core component all customers that update would benefit from it. 

> So the question remains, what are we doing until then? One approach could be to only allow extending filters for now and once we have CIF-2916 we additionally allow to replace them, but without passing the identifier.

I don't see why we should use two methods to do one thing, this is in generally harder to get (when to use what?).

With `Function<>` the API would look the same, and be fluent as well

```
productRetriever.extendProductFilterWith(f -> f
    .setCustomFilter("eco-friendly", new FilterEqualTypeInput()
        .setEq("yes")));
```

```
productRetriever.extendProductFilterWith(f -> new ProductAttributeFilterInput()
   .set... 
   .setCustomFilter("eco-friendly", new FilterEqualTypeInput()
        .setEq("yes")));
```

Buuhuu: And then what? If we go for the `Consumer<>` changing that in the future will be breaking change. So it would force use to go with 2 methods for one thing, which leads again to: 

> I don't see why we should use two methods to do one thing, this is in generally harder to get (when to use what?).

LSantha (approved):

PR #943, opened 2022-07-15
CIF-2916: add UrlProvider product and category filter hooks

Buuhuu:
## Description

Currently the `UrlProvider` has to do a query when a component wants to get the product identifier (sku) or category (uid) and the requested value cannot be parsed from the url by the url format. Usually this identifier is later used by the components to query for either the product (e.g product detail, breadcrumb, related products) or some category data.

This change moves this one query done by the `UrlProvider` in some cases to further reduce the number of graphql request per page. It does so by extending the retrieves with a product / category filter hook obtained from the `UrlProvider`. These hooks have access to the parameters the url format can parse from the url and construct the filter input accordingly. 

Furthermore the request-local caching implemented in the `MagentoGraphqlClientImpl` got extended to respond also for partially matching queries (a very simple form of a higher level graphql cache). This helps for example any commerce content fragment or commerce experience fragment on a product detail page, as they can reuse the response from the product detail query (that contains the sku they get from the `UrlProvider#getProductIdentifier()`).

In Venia that means:
- PDP: 
-- one query less, Commerce Experience Fragment reuses response of the product detail component
-- remaining queries are: navigation, breadcrumb, product details
- PLP:
-- one query less, Products query still uses the uid from the category query (may be further optimised by doing the category query and products query together in a single request)
-- remaining queries are: navigation, breadcrumb, product list (introspection query, custom attribute metadata query, category and products query)

Some details about the changes:
- Changing the Product and ProductList component was tricky as both show the placeholder data on author als when the product/category could not be found in the commerce backend. With the new approach this is only possible after doing the actual queries and so just replacing the retrievers in this case on author doesn't work anymore. Instead both components now handle placeholder data a little different 
- wrt to the fuzzy local cache hit in the MagentoGraphqlClientImpl: I changed all the products queries in order to always query the sku, url_key first so that the queries actually match with the one the url provider would do. 

## Related Issue

CIF-2916
https://github.com/adobe/aem-core-cif-components/pull/938

## Motivation and Context

Each graphql query that is executed during page rendering comes on a cost, essentially blocking the rendering until the response is received. The query done by the url provider usually is fast but the unnecessary network round trip can still be avoided. 

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:


examples/bundle/pom.xml
```diff
@@ -138,28 +138,6 @@
                                     <goal>restore-instrumented-classes</goal>
                                 </goals>
                             </execution>
```
LSantha, line 140: Wouldn't be safer to keep this?

examples/ui.apps/pom.xml
```diff
@@ -74,6 +74,11 @@
                         <createdBy>${vault.package.company}</createdBy>
                         <cloudManagerTarget>none</cloudManagerTarget>
                     </properties>
+                    <validatorsSettings>
```
LSantha, line 77: Any reason for disabling it?

LSantha (approved):

examples/bundle/pom.xml
Buuhuu: The reason why I removed it is that we have a tight coupling of these tests with the core implementation. We do have the codecov check which makes this redundant, but the codecov dashboard gives good insights what has to be covered.

PR #946, opened 2022-08-01
CIF-2842 - Extend List component to list pages by product or category

LSantha:
* initial implementation of commerce list component
 * added unit tets
 * updated examples and documentation

## Related Issue

CIF-2842

## How Has This Been Tested?

Unit test, UI test, IT, manually.

## Types of changes

## Checklist:


LSantha:
The tests will pass after CIF add-on release.

bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/models/v1/list/CommerceListImpl.java
```diff
@@ -0,0 +1,232 @@
...
+                : siteStructure.isCategoryPage(currentPage) ? urlProvider.getCategoryIdentifier(request) : null;
+            if (StringUtils.isNotBlank(categoryUid)) {
+                contentQuery = associatedContentService.listCategoryContentPages(resourceResolver,
+                    PageParams.of(categoryUid).path(siteStructure.getLandingPage().getPath()));
+            } else {
+                LOGGER.warn("Category identifier missing for Commerce List at " + resource.getPath());
+            }
+        }
+
+        if (contentQuery == null) {
```
Buuhuu, line 134: Iiuc this causes the component to not support any of the WCM List's behaviours right? Shouldn't we implement a fallback here? Otherwise customers would end up with two list components, one for the wcm case and one for the cif case.

Ideally we could just create the super type model without providing any additional parameters in this case. This would make the Commerce List component a drop in replacement for the wcm list.

LSantha: Actually the wcmList is already initialized via injection, so nothing to do there. 
My intention is to provide a drop in replacement  for WCM List.

Buuhuu (approved):

PR #966, opened 2022-10-18
CIF-2982 - Support product filter hooks in ProductList component

herzog31:
## Description

* Made `CategoryRetriever` from `ProductList` component public.
* Extended `CategoryRetriever` with product filter extension. This is for setting and getting only, not actually used by this retriever.
* Added unit test.

## How Has This Been Tested?

* Unit test
* Proxy component in Venia

## Types of changes

## Checklist:


Buuhuu:
With this approach we now have two ways to extend the query of the product list component: by using the productQueryHook of the `ProductList#getCategoryRetriever()` and by using the newly introduced filter hook.

Wouldn't it make more sense to add the product filter hook to the AbstractCategoryRetriever? That way the extension pattern remains the same. Also that the SearchResults and the ProductList use the same queries is an implementation detail, it does not mean that any ProductCollection would allow to do that.

herzog31:
@Buuhuu I think we have a few consistency issues here:

- `SearchResults` doesn't use any retriever, so we cannot use the same pattern here as with the `ProductList`.
- `AbstractCategoryRetriever` is used in the context of a category query. While extending category, category filter and products make sense here, extending the product filter does not. This retriever would not be using it for anything, but we would only have an external consumer.
- I agree that moving the extensions to `ProductCollection` might not be the best place, since as you said, we cannot expect any subclass to allow the same grade of extensibility.

I don't have a good answer here, since frankly the `SearchResultsService` is a mess.

Buuhuu (approved):

PR #988, opened 2023-09-08
aep extension bug fixes (commerce ML)

aniham:
- upgrade to latest version of collector
- generate search results view event
- remove context from search events (should be picked up from data layer)
- return facets as list instead of null (schema validation failures)


bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/storefrontcontext/SearchResultsStorefrontContextImpl.java
```diff
@@ -80,7 +81,7 @@ public int getPerPage() {
     @Override
     public List<SearchFacet> getFacets() {
         // TODO: return the facets as we have them available in our result set
-        return null;
+        return new ArrayList<SearchFacet>();
```
Buuhuu, line 84: Why is that needed? This is a change of behaviour in the API. Imaging, customer code checks the return value for `null`. That would not work anymore after that change.

aniham: @Buuhuu this method is only referenced when setting StorefrontContext. I don't believe any custom code could or would be using it. Moreover, as I mentioned in the PR summary, this now breaks how events are received for data collection. So. It looks like it was originally implemented wrong and never caught as a bug in testing.

Buuhuu (approved):

herzog31:
Replaced by #991

PR #994, opened 2023-11-27
SITES-17412: Log WARN when getting duplicate id from Magento GraphQL API

hrayrpapikyan:
Log WARN when getting duplicate id from Magento GraphQL API
## Description

Log WARN when getting duplicate id from Magento GraphQL API

## Related Issue
https://jira.corp.adobe.com/browse/SITES-17412

## Motivation and Context

There exists a way to get duplicate data from Magento GraphQL API, this possibility occurs when customers add data to their system and the indexers do not complete before data is retrieved from the system.

 

Do It Best was affected by this and the AEM Support Org spent weeks debugging (because it was also a very intermittent issue). Eventually, T3 deployed a custom build of the Category Feature List showcasing the GraphQL query and we discovered that the categoryIds were unique. After engaging and reviewing our findings with SITES engineering we agreed to log a ticket with Magneto Engineering which then Magento ack'd that this can occur.

 

The CIF should log a WARN in the logs to help inform the customer that they need to fix the data in their Magento instance so that their site is not showcasing incorrect data. 

 

In the situation where the customer has [^components .content.doitbest.us.en.global.homepage.jcr-content.root.json] we should check that the resulting GraphQL response does not contain duplicate categoryId or else this would be an issue.
## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/retriever/AbstractCategoriesRetriever.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/models/retriever/AbstractCategoriesRetriever.java
@@ -17,11 +17,15 @@ package com.adobe.cq.commerce.core.components.models.retriever;
 
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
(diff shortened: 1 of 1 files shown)


LSantha (approved):

LSantha:
@hrayrpapikyan , need to check the tests! The codecov check would usually fail for this PR because the new code doesn't have any coverage.

PR #1007, opened 2024-06-05
Adding set error to search result retriever

pat-lego:
## Description

Adding setter for error when error is not empty 

Internal -> CIF-3020


LSantha:
Good point @pat-lego , I have incorporated this fix in the PR https://github.com/adobe/aem-core-cif-components/pull/1008 together with other fixes. I'm closing this PR.

PR #1047, opened 2025-09-29
SITES-34368: Updates the magento graphql version

alwinjoseph02:


LSantha:
Sorry @alwinjoseph02 , just realized that we don't need this specific PR because we are not using the new API in this project yet..

PR #1071, opened 2026-05-07
SITES-43978 Publish returns 404

Satish-Pothabathula:
Summary
Fix a regression where requests resolving to jcr:content (or any descendant) on CIF catalog pages could incorrectly return HTTP 404 from CatalogPageNotFoundFilter.

Problem
CatalogPageNotFoundFilter is intended to validate page-level rendering requests that carry product/category context in the URL (for example, a PDP request with a suffix containing the SKU/url_key).

However, the filter was also invoked for raw content-tree reads such as:

.../product-page/jcr:content.json
.../product-page/jcr:content/root.json
(and similarly for any subcomponent under jcr:content)
Those requests do not carry SKU/category context, so the filter could end up adapting product/category models without the required URL context and return 404, even though the content exists.

Fix
Add an early guard at the top of doFilter() to short-circuit when the request resource path is jcr:content or a descendant:

If the resource path ends with "/jcr:content" or contains "/jcr:content/", the filter simply calls filterChain.doFilter(...) and returns.
This ensures the filter only validates page-level requests, and does not interfere with raw content reads or subcomponent renders under jcr:content.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/servlets/CatalogPageNotFoundFilter.java
+++ b/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/servlets/CatalogPageNotFoundFilter.java
@@ -25,6 +25,7 @@ import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletResponse;
 
+import org.apache.jackrabbit.JcrConstants;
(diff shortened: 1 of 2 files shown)


LSantha (approved):

LSantha:
Solution available in 2.18.4 release of CIF components.
Release notes: https://github.com/adobe/aem-core-cif-components/releases/tag/core-cif-components-reactor-2.18.4

PR #1080, opened 2026-07-16
SITES-46557 - Add GitHub Actions CI/release workflows

alwinjoseph02:


.github/scripts/deploy-queries.js
```diff
@@ -0,0 +1,87 @@
...
+    const tagName = `components-queries-${version}`;
+    ci.gitImpersonate('GitHub Actions', 'no-reply@adobe.com', () => {
+        if (doCommit) {
+            // Commit and push changes
+            ci.sh(`git commit -m "releng - Update Queries for CIF Core Components v${version}"`);
+            ci.sh(`git push --set-upstream origin master`);
+        }
+        // Tag latest commit with component release
+        ci.sh(`git tag ${tagName} HEAD`);
+        ci.sh(`git push origin ${tagName}`);
```
Satish-Pothabathula, line 67: If this runs twice for the same version (e.g. a release retry), git tag ${tagName} fails because the tag already exists and aborts the step. Consider git tag -f or checking for the tag first so retries are idempotent.

alwinjoseph02: The same behaviour exists in circleci too. And from my understanding it shouldn't run on the same tag again will create other issues

.github/workflows/ci.yml
```diff
@@ -0,0 +1,496 @@
...
+        if: always()
+        uses: actions/upload-artifact@v6
+        with:
+          name: karma-junit
+          path: ui.apps/karma-junit
+          if-no-files-found: ignore
+
+      # if: failure() (not always()): skip this step entirely on a passing run - only
+      # run it (and thus only create a check run) when there's an actual failure.
+      - name: Publish test results
```
Satish-Pothabathula, line 149: CircleCI never set an explicit timeout in `config.yml` — it relied on the built-in
per-step default of **10 minutes of no output** (`no_output_timeout: 10m`), which
automatically killed a hung step (e.g. an AEM startup that never comes up).

but GitHub Actions has **no no-output timeout** and defaults to a **6-hour** job limit,
so that safety net is gone after the migration. A hung `test-aem` job would sit for
6h before failing.

Suggest adding `timeout-minutes` to each job (especially the AEM integration/selenium
jobs in `ci.yml`) to restore the bound CircleCI gave us for free, e.g.:

    jobs:
      test-aem:
        runs-on: ubuntu-latest
        timeout-minutes: 30

